PR #2, opened 2022-06-01
feat: add support to seencheck from HQ

NGTmeaty:


internal/pkg/crawl/capture.go
```diff
@@ -150,6 +150,18 @@ func (c *Crawl) captureAsset(item *frontier.Item, cookies []*http.Cookie) error
...
 	}
 
+	if c.UseHQ {
+		found, err := c.hqSeencheck(item.URL)
+		if err != nil {
+			// potentially allow failure here, since this is only seencheck and HQ could have high load or another issue
+			return err
+		}
+		if found {
+			// Since the asset is already seenchecked in HQ, we don't need to crawl it, as well as having been added to the local seencheck above, line 150.
```
CorentinB, line 160: Maybe not a very good idea to refer to lines, as line numbers move when adding code over time.

internal/pkg/crawl/capture.go
```diff
@@ -150,6 +150,18 @@ func (c *Crawl) captureAsset(item *frontier.Item, cookies []*http.Cookie) error
 		c.Frontier.Seencheck.Seen(hash, item.Type)
 	}
 
+	if c.UseHQ {
+		found, err := c.hqSeencheck(item.URL)
+		if err != nil {
+			// potentially allow failure here, since this is only seencheck and HQ could have high load or another issue
```
CorentinB, line 156: Yes, I think we should just crawl if HQ fails to answer us. Better to be duplicating than loosing stuff.

internal/pkg/crawl/hq.go
```diff
@@ -115,3 +115,31 @@ func (c *Crawl) hqFinisher() {
 
 	}
 }
+
+func (c *Crawl) hqSeencheck(URL *url.URL) (bool, error) {
```
CorentinB, line 119: Technically, as per Go specs, `hq` should always be `HQ`.

CorentinB (changes requested):

PR #4, opened 2022-06-03
Add: batch HQ seencheck

CorentinB:


internal/pkg/crawl/capture.go
```diff
@@ -349,6 +349,23 @@ func (c *Crawl) Capture(item *frontier.Item) {
 		return
 	}
 
+	// If we didn't find any assets, let's stop here
+	if len(assets) == 0 {
+		return
+	}
+
+	// If we use HQ, we seencheck the assets
+	if c.UseHQ {
```
NGTmeaty, line 358: I'm not a huge fan of using the remote HQv3 seencheck first and would much rather check the local, (most likely) significantly faster, seencheck first, then the remote HQv3?

CorentinB: Understandable. It's done here because it prepares the batch of assets to be sent to the goroutines just after, in which local seencheck happens.

Makes me think, shouldn't we simply disable local seencheck when HQ is being used? In which case the `if c.Seencheck...` condition below should contain `&& !c.UseHQ`. What do you think?

NGTmeaty: After talking on Slack, we've resolved this. In a later commit I will fully fix the behavior here.

PR #9, opened 2022-06-07
Filter out outlinks that being with a # (fragment or hash URL)

NGTmeaty:
Fragment or hash URLs almost always end up on the exact same page, especially ones without a full URL.

URLs like: "#main" or "#content" would be filtered out.

Important note: this won't filter out outlinks that contain fragment URLs if they are full URLs. (Maybe stripping out the fragments to allow proper seenchecking could make more sense in a future PR?)


CorentinB:
Thanks! Maybe add a bit of comment in the code itself to explain that  if? Beside that, all good.

PR #19, opened 2022-06-20
Code cleanup and some improvements

NGTmeaty:
I made some improvements to retry on errors to ensure they don't immediately retry and cleaned up capture.go from the legacy markTempFileDone which isn't used.


CorentinB:
Looks good to me, thanks!

PR #28, opened 2022-06-30
Add real version number

NGTmeaty:
Adds commit hash to warcinfo record, this is currently returning the entire hash, maybe we should return the first 7 characters like GitHub, which retains enough information to figure out exactly what code was writing to WARC? 

In warcinfo:
Current commit: `software: Zeno ae02086213b09b0a85388321596a99a1995e6ce2-true`
Previously: `software: Zeno`
Proposed additional, possible, change: `software: Zeno ae02086-true`
(and if nothing was modified in the local git tree: `software: Zeno ae02086`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 jobs/*
+Zeno
--- a/cmd/all/all.go
+++ b/cmd/all/all.go
@@ -2,4 +2,5 @@ package all
 
 import (
 	_ "github.com/CorentinB/Zeno/cmd/get"
+	_ "github.com/CorentinB/Zeno/cmd/version"
 )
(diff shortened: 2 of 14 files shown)


CorentinB:
Great addition! Would be nice to add a `version` command!
Rclone also adds more data like:
```
corentin@excalibur:~/Documents/work » rclone version
rclone v1.58.1
- os/version: darwin 12.4 (64 bit)
- os/kernel: 21.5.0 (arm64)
- os/type: darwin
- os/arch: arm64
- go/version: go1.18.1
- go/linking: dynamic
- go/tags: none
```

I like that, maybe we could try something similar.

NGTmeaty:
LGTM!

PR #62, opened 2024-05-15
Allow crawl space threshold to be set on CLI, report space avail

### LINKED ISSUE #61: Allow free space threshold to be customizable
I built Zeno from source ([687b5d5](https://github.com/internetarchive/Zeno/commit/687b5d5982be433206b03022d0a03dc0a1227501)) and ran `Zeno get url` only be told I did not have enough space. It would be great if (1) this value was customizable (it appears to be hard-coded to 20 GB) and/or (2) the amount of space needed was reported to the user.

machawk1:
Closes #61

@CorentinB  Please note, I'm not proficient in Go, so feedback is welcomed and edit at will.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -169,6 +169,13 @@ var GlobalFlags = []cli.Flag{
 		Usage:       "Number of seconds until the crawl will automatically panic itself. Default to crawl-time-limit + (crawl-time-limit / 10)",
 		Destination: &config.App.Flags.MaxCrawlTimeLimit,
 	},
+	&cli.IntFlag{
+		Name:        "min-space-required",
+		Aliases:     []string{"msr"},
+		Value:       20,
+		Usage:       "Minimum space required to start crawl",
+		Destination: &config.App.Flags.MinSpaceRequired,
+	},
 
 	// Proxy flags
 	&cli.StringFlag{
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -173,7 +173,7 @@ var GlobalFlags = []cli.Flag{
 		Name:        "min-space-required",
 		Aliases:     []string{"msr"},
 		Value:       20,
-		Usage:       "Minimum space required to start crawl",
+		Usage:       "Minimum space (GB) required to start crawl",
 		Destination: &config.App.Flags.MinSpaceRequired,
 	},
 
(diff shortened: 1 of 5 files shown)


cmd/cmd.go
```diff
@@ -169,6 +169,13 @@ var GlobalFlags = []cli.Flag{
 		Usage:       "Number of seconds until the crawl will automatically panic itself. Default to crawl-time-limit + (crawl-time-limit / 10)",
 		Destination: &config.App.Flags.MaxCrawlTimeLimit,
 	},
+	&cli.IntFlag{
+		Name:        "min-space-required",
+		Aliases:     []string{"msr"},
+		Value:       20,
+		Usage:       "Minimum space required to start crawl",
+		Destination: &config.App.Flags.MinSpaceRequired,
+	},
```
CorentinB, line 178: Might be good to specify that the minimu space required is in GB.

internal/pkg/crawl/utils.go
```diff
@@ -52,8 +53,10 @@ func (c *Crawl) checkIncludedHosts(host string) bool {
 
 func (c *Crawl) handleCrawlPause() {
 	for {
-		if float64(utils.GetFreeDiskSpace(c.JobPath).Avail)/float64(GB) <= 20 {
-			logrus.Errorln("Not enough disk space. Please free some space and restart the crawler.")
+		spaceLeft := float64(utils.GetFreeDiskSpace(c.JobPath).Avail) / float64(GB)
+		if spaceLeft <= float64(c.MinSpaceRequired) {
+			logrus.Errorln(fmt.Sprintf("Not enough disk space: %d GB required, %f GB available", c.MinSpaceRequired, spaceLeft))
+			logrus.Errorln("Please free some space and restart the crawler.")
```
CorentinB, line 59: Only freeing some space is required, the crawler will unpause itself when it has enough disk space.

CorentinB, line 59: Maybe use a single logrus call?

CorentinB (changes requested):
I made few suggestions, but nothing big mostly cosmetic :)

CorentinB:
Thanks for your contribution @machawk1, to be noted: **WARC writing is async**. The WARC writing queue is displayed when using Zeno with --live-stats, one edge case that is not handled by my code currently is when disk space is too low so the crawl pause, but at the same time there is enough in the WARC writing queue that the remaining space can be eaten up entirely.

Because the WARC writing isn't paused by this pause mechanism here. I mean the WARC writing that was already in the WARC writing queue. I hope it's clear.

**Not saying you have to do anything about it here**, and solving this would require some additional synchronization between Zeno and the underlying [warc library](https://github.com/CorentinB/warc), but I thought it was worth mentioning. :)

CorentinB (approved):

CorentinB:
Thanks!  Really sorry for the delay here..

PR #63, opened 2024-06-24
Add state management for workers

equals215:
- Workers are now bound to a an operator-like structure, their state is visible and manipulated through that object
- Workers state is visible using API queries


CorentinB (approved):

equals215:
Not working, bringing the changes on the original repo

PR #66, opened 2024-06-29
Give Zeno the logging it deserves

equals215:
Zeno logging is chaotic and need to be standardized.  
The current standard for logging is `slog` as it was recently introduced in the Go standard library.  

- `multiHandler`  implements `slog.Handler` interface with support for multiple outputs, currently fixed with stdout and file output
- file logging is done in json and stdout logging is in text format, both outputs' logging level can be configured

TODO :

- [x] replicate `utils.SetupLogging()` behaviour in `log`
  - [x] ElasticSearch logging should be handled by `log`
  - [x] Index rotation every day


equals215:
yeah it works, elasticsearch log forwarder with slog implementation (sick huh?)

Screenshot 2024-06-29 at 9 26 34 PM

equals215:
@CorentinB your turn!

equals215:
still need to add the capacity to specify log file path, will do in the upcoming CLI rewrite PR

cmd/utils.go
```diff
@@ -1,31 +1,73 @@
...
+		RotateElasticSearchIndex: true,
+		ElasticsearchConfig:      elasticSearchConfig,
+	})
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	c.Log = customLogger
+
+	go func() {
```
CorentinB, line 54: Not a huge impact but it could be slightly better to handle the closing of this goroutine when Zeno is stopped. (using a waitgroup to make sure it properly finished before exiting Zeno etc)

config/config.go
```diff
@@ -62,9 +62,12 @@ type Flags struct {
 	DisableAssetsCapture bool
 	CertValidation       bool
 
-	CloudflareStream bool
-	ElasticSearchURL string
-	ExcludedStrings  cli.StringSlice
+	CloudflareStream         bool
```
CorentinB, line 65: Not your fault but I just noticed that this bool is never used, therefore completely useless, you can remove it.

internal/pkg/log/elasticsearch.go
```diff
@@ -0,0 +1,148 @@
...
+		if strings.Contains(err.Error(), "EOF") {
+			return fmt.Errorf("error creating index: received EOF from Elasticsearch, is the server running? check your ES logs for more information")
+		}
+		return fmt.Errorf("error creating index: %w", err)
+	}
+	defer res.Body.Close()
+
+	if res.IsError() {
+		// If the index already exists, that's okay
+		if strings.Contains(res.String(), "resource_already_exists_exception") {
```
CorentinB, line 141: Are you sure the ES pkg doesn't have a variable defining the error? (could be slightly better than comparing the text)

internal/pkg/log/rotate.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"fmt"
+	"log/slog"
+	"os"
+	"time"
+)
+
+// ... (previous Logger, multiHandler, logEntry, ElasticsearchHandler definitions remain the same)
+
+type rotateableHandler interface {
```
CorentinB, line 12: rotatable*

CorentinB (changes requested):
Looks VERY good to me. Once the few small changes requested are done, it's good to merge. Thanks!

internal/pkg/log/elasticsearch.go
equals215: not that I can find

internal/pkg/log/rotate.go
equals215: indeed

equals215 (commented):
done

PR #75, opened 2024-07-10
Remove Gin dependency

CorentinB:
This PR replace the usage of Gin with the standard net/http lib.


CorentinB:
@equals215 it's out of the scope of this PR but I wondered if `/worker/0` should be modified to be `/workers/0`, in order to be consistent with the `/workers` endpoint. I understand why you named it `worker` and not `workers` though, but I felt like it should be named like the `/workers` endpoint anyway. What do you think?

PR #78, opened 2024-07-12
Rewriting the queue

CorentinB:
The goal of this work is to replace the current queuing mechanism that use `github.com/beeker1121/goque` with a custom piece of code that exactly fit our needs and on which we will have much better control.

Almost all queuing components have been rewritten except the host crawl rate limiter, I commented out the relevant parts with `// TODO: re-implement host limitation`, @NGTmeaty if you can look into that part as you were the original author, to make it fit in the current design, it would be nice. Having it as a separate package would also be great.

I also separated the queue and seencheck packages, instead of having 1 big "frontier" package. FWIW, I'm also moving away calling it "frontier", it has only caused misunderstanding from the people I explained it too. (it was originally named like that to follow Heritrix3 path.. but let's clean things up and not use that term anymore)

Finally, I added a big list of statistics and infos about the queue that are available through the API at `/queue`.

Suggestions & code-review are welcome and well-needed. :)


internal/pkg/crawl/crawl.go
```diff
@@ -147,8 +150,8 @@ func (c *Crawl) Start() (err error) {
 	// to exit Zeno, like CTRL+C
 	go c.setupCloseHandler()
 
-	// Initialize the frontier
-	frontierLoggingChan := make(chan *frontier.FrontierLogMessage, 10)
+	// Initialize the frontier components
```
equals215, line 153: if the package was renamed queue, maybe the `frontier` naming isn't relevant anymore?

internal/pkg/crawl/crawl.go
```diff
@@ -147,8 +150,8 @@ func (c *Crawl) Start() (err error) {
...
+	frontierLoggingChan := make(chan *queue.LogMessage, 10)
```
equals215, line 154: same here^

internal/pkg/queue/dequeue.go
```diff
@@ -0,0 +1,101 @@
...
+		var item Item
+		err = q.queueDecoder.Decode(&item)
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode item: %w", err)
+		}
+
+		// Move to next host
+		q.currentHost = (q.currentHost + 1) % len(q.hostOrder)
+
+		// Update stats
```
equals215, line 76: can stats update logic be in another function? I feel that it doesn't belong in the `Dequeue()` func

internal/pkg/queue/dequeue_test.go
equals215, line 1: très bon les tests, on comprend bien plus facilement le fonctionnement du pkg

internal/pkg/queue/enqueue.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// Update host index and order
+	q.hostMutex.Lock()
+	if _, exists := q.hostIndex[item.Host]; !exists {
+		q.hostOrder = append(q.hostOrder, item.Host)
+	}
+	q.hostIndex[item.Host] = append(q.hostIndex[item.Host], uint64(startPos))
+	q.hostMutex.Unlock()
+
+	// Update stats
```
equals215, line 37: same remark as `Dequeue()`

equals215 (approved):
lgtm, comments are mostly suggestions
good job Coco it looks really nice 😊

(i'm already crying blood from thinking I'll need to merge this to my stale PR)

internal/pkg/queue/dequeue.go
```diff
@@ -0,0 +1,72 @@
...
+		err = q.saveMetadata()
+		if err != nil {
+			return nil, fmt.Errorf("failed to save metadata: %w", err)
+		}
+
+		return &item, nil
+	}
+
+	// If we've checked all hosts and found no items, loop back to wait again
+	return q.Dequeue()
```
equals215, line 71: you really sure you wanna go down the recursive path? what happens if a function calls Dequeue() when the queue is empty and mistakenly stackoverflow?

equals215:
aligned queue structures to save space
```
Zeno/internal/pkg/queue/metadata.go:17:15: 16 bytes saved: struct with 224 pointer bytes could be 208
Zeno/internal/pkg/queue/metadata.go:63:14: 16 bytes saved: struct with 224 pointer bytes could be 208
Zeno/internal/pkg/queue/queue.go:30:29: 80 bytes saved: struct with 360 pointer bytes could be 280
Zeno/internal/pkg/queue/queue.go:52:11: 32 bytes saved: struct with 104 pointer bytes could be 72
Zeno/internal/pkg/queue/stats.go:8:17: 64 bytes saved: struct with 184 pointer bytes could be 120
```

equals215:
As Corentin is a complete clown when it comes to using git, here's the commands to pull the branch if git yells at you when you're trying to pull changes :
```bash
git checkout queue
git fetch origin
git reset --hard origin/queue
```

CorentinB:
Things left to fix/do:

- [ ] Add compaction
- [ ] Switch from Gob to Protobuf for metadata
- [ ] Find a way to save metadata more efficiently

internal/pkg/queue/enqueue.go
```diff
@@ -0,0 +1,52 @@
...
+
+	if q.closed {
+		return ErrQueueClosed
+	}
+
+	q.mutex.Lock()
+	defer q.mutex.Unlock()
+
+	// Find free position
+	startPos, err := q.queueFile.Seek(0, io.SeekEnd)
```
equals215, line 22: here to fill the gaps

equals215: turning it into an issue

equals215:
Yeah, that's pretty much it

equals215:
TODO : 
- [ ] make sure all index temp files are created in the same dir to avoid cross-disk issues like this one
```log
2024/07/25 15:40:21 ERROR Periodic dump failed error="failed to perform periodic dump: failed to rename temp file to index file: rename /tmp/index_dump_3688963110 jobs/queue/index: invalid cross-device link"
```
- [ ] rotate index gob encoder/decoder after a successful dump

equals215:
TODO : Is the queue still rotating hosts properly? If not we can implement an atomic uint64 that increases with every dequeue and apply a modulo of the `len(hosts)` which will rotate the hosts on Dequeue() the most efficient way possible, can even be saved as state ultra easily and if it overflows we mostly don't care

equals215:
TODO :
- [ ] Send seeds directly to the workers when available using a `sync.Cond` effectively bypassing the local queue to gain performances
- [ ] Implement a space recollection index to write items over dequeued but not deleted items

internal/pkg/queue/enqueue.go
```diff
@@ -0,0 +1,133 @@
...
+
+			itemsChan <- &msg{
+				bytes: b,
+				item:  i,
+			}
+		}(item, wg)
+	}
+
+	wg.Wait()
+	close(itemsChan)
```
equals215, line 64: @CorentinB what in the banana ham jelly bagel sandwich is that, you close a channel BEFORE reading it? and you wonder why we don't crawl much

CorentinB: Question is: how does it even work right now

CorentinB: @equals215 it's perfectly fine to do that, the channel is closed but the elements are still in it and can be consumed, the range just after is working fine

yzqzss:
@equals215 

image

time=xxxxxxxxx level=WARN msg="[WORKERS] Timeout reached. 2 workers still running"

Workers still get stuck in the second inner for.
revert to if statement?

CorentinB:
@yzqzss I think `--batch-write-WAL` should be `--batch-write-wal`, **but** any occurence of `WAL` in the code in variables (such as `walPath` for example) should be capitalized. (as per Go variable naming convention)

yzqzss:
> @yzqzss I think `--batch-write-WAL` should be `--batch-write-wal`, **but** any occurence of `WAL` in the code in variables (such as `walPath` for example) should be capitalized. (as per Go variable naming convention)

yeap, I think `WalIoPercent` is a bit stupid too. feel free to change them yourself. :)

equals215:
finally!

internal/pkg/queue/queue.go
```diff
@@ -37,6 +27,8 @@ type PersistentGroupedQueue struct {
 	mutex           sync.RWMutex
 	statsMutex      sync.RWMutex
 	closed          bool
+
+	logger *slog.Logger
```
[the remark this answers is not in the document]
yzqzss: #108

PR #79, opened 2024-07-15
Fix crash when --api is not set

NGTmeaty:
Prometheus was always expecting to be present but API is not always set when crawls are ran, resulting in Prometheus not being correctly defined. This resolves that issue.


CorentinB:
Thanks!

PR #80, opened 2024-07-15
Fix workers not stopping properly and temp fix for workers hanging unexpectedly

equals215:
### !!Not a definitive fix!!

- Corrected an unexpected behaviour where workers got killed while in a retry loop
- Lowered the `max-retry` default from 20 to 5
- Logging the retries to enhance debugging
- Added a `lastAction` field on Worker status struct to enhance logging and debugging


equals215:
TODO: make the state handling + deletion of workers goroutine safe
get rid of array and use a map to store workers

PR #95, opened 2024-07-24
Implement linkheader parsing

### LINKED ISSUE #85: Replace github.com/tomnomnom/linkheader with stdlib
github.com/tomnomnom/linkheader is being used to parse Link HTTP headers, pretty sure we can replace it with standard lib-only code.

Whatever code we add, we need unit test for it. (we don't have unit tests for that right now)

HarshNarayanJha:
This PR adds a Parse function for linkheader parsing, therefore removing the dependency on `github.com/tomnomnom/linkheader`. This also includes writing unit tests for the function

Closes #85

### DIFF
--- a/internal/pkg/crawl/link_header.go
+++ b/internal/pkg/crawl/link_header.go
@@ -1,7 +1,6 @@
 package crawl
 
 import (
-	"regexp"
 	"strings"
 )
 
@@ -21,17 +20,23 @@ type Link struct {
 func Parse(link string) []Link {
 	var links []Link
 
-	urlRegex := regexp.MustCompile(`^<(\S+)>$`)
-	// var linkRegex = regexp.MustCompile(`^<(\S+)>(; (\S+)=(\S+))*$`)
-
 	for _, link := range strings.Split(link, ", ") {
-		parts := strings.Split(link, "; ")
-		match := urlRegex.FindStringSubmatch(parts[0])
-		url := match[1]
+		parts := strings.Split(link, ";")
+		if len(parts) < 1 {
+			// Malformed input, somehow we didn't get atleast one part
+			continue
+		}
+
+		url := strings.TrimSpace(strings.Trim(parts[0], "<>"))
 		rel := ""
 
 		for _, attrs := range parts[1:] {
 			key, value := ParseAttr(attrs)
+			if key == "" {
+				// Malformed input, somehow the key is nothing
+				continue
+			}
+
 			if key == "rel" {
 				rel = value
 				break
@@ -45,8 +50,14 @@ func Parse(link string) []Link {
 
 // Parse a single attribute key value pair and return it
 func ParseAttr(attrs string) (key, value string) {
-	attrRegex := regexp.MustCompile(`^(\S+)=\"(\S+)\"$`)
-	match := attrRegex.FindStringSubmatch(attrs)
+	kv := strings.SplitN(attrs, "=", 2)
 
-	return match[1], match[2]
+	if len(kv) != 2 {
+		return "", ""
+	}
+
+	key = strings.TrimSpace(kv[0])
+	value = strings.TrimSpace(strings.Trim(kv[1], "\""))
+
+	return key, value
 }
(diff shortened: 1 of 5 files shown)


HarshNarayanJha:
Run the tests using `go test -v $(go list ./... | grep crawl) `, or run `go test -v` from the `crawl` directory.
Don't know why, but go doesn't want to run all the tests from the project root.

HarshNarayanJha:
Is the implementation alright? Should I go ahead and replace the usages in `capture.go`?

internal/pkg/crawl/link_header.go
```diff
@@ -0,0 +1,52 @@
...
+func Parse(link string) []Link {
+	var links []Link
+
+	urlRegex := regexp.MustCompile(`^<(\S+)>$`)
+	// var linkRegex = regexp.MustCompile(`^<(\S+)>(; (\S+)=(\S+))*$`)
+
+	for _, link := range strings.Split(link, ", ") {
+		parts := strings.Split(link, "; ")
+		match := urlRegex.FindStringSubmatch(parts[0])
+		url := match[1]
```
willmhowes, line 30: If there is no match and the value of match is `nil`, then the program will halt due to trying to access an index on a `nil` value. The function needs to be written to handle malformed inputs.

willmhowes: There should also be a test added to demonstrate the handling of malformed input

internal/pkg/crawl/link_header.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Parse parses a raw Link header in the form:
+//
+//	<url1>; rel="what", <url2>; rel="any"; another="yes", <url3>; rel="thing"
+//
+// returning a slice of Link structs
+// Each of these are separated by a `, ` and the in turn by a `; `, with the first always being the url, and the remaining the key-val pairs
+// See: https://simon-frey.com/blog/link-header/, https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link
+func Parse(link string) []Link {
```
willmhowes, line 21: Because regular expressions can be CPU intensive, can you implement the `Parse` function without them? One way to do this would be using the Go strings `Split` and `Trim` functions to breakdown the input string by the separator characters defined in the [Link Header specification](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link), which are `;` and `,` and `<>`. You already started this approach using `Split` on `,` so great job there!

willmhowes (changes requested):
Great work so far! I've left two pieces of feedback as comments, please let me know if you have any questions and we can take it from there. Cheers!

internal/pkg/crawl/link_header.go
HarshNarayanJha: Should I completely ignore malformed inputs, or just try to parse url out of it?
e.g.

 - `<>;` this can completely be ignored
 
But url can be extracted from these

 - `https://test.com>; rel=;`
 - `https://another.com;;;;`

CorentinB: If you can try and succeed at parsing URLs out of malformed inputs, no harm in trying to. Just make sure you are not parsing false positives 99% of the time. (1% of false positives, _just making up numbers here_, is fine..)

internal/pkg/crawl/link_header.go
HarshNarayanJha: Test Added

internal/pkg/crawl/link_header.go
HarshNarayanJha: Okay! Done

HarshNarayanJha:
Replaced regexep with Trim and Split, and now it also tries to parse URL out of malformed inputs, otherwise just returns empty slice

HarshNarayanJha:
Removed linkheader dependency and updated `capture.go` to use the new `Parse` function

willmhowes (approved):
LGTM, nice work!

HarshNarayanJha:
Something's preventing the merge? Did I miss anything?

PR #97, opened 2024-07-25
create url_string_test.go

willmhowes:
Pretty straightforward testing for `URLToString` function. But I'm still not quite sure what the following logic in `URLToString` does so I wasn't able to write a test for it ([link](https://github.com/internetarchive/Zeno/blob/main/internal/pkg/utils/url_string.go#L28)):

```golang
if strings.Contains(tempHost, ":") && !(strings.HasPrefix(tempHost, "[") && strings.HasSuffix(tempHost, "]")) {
	tempHost = "[" + tempHost + "]"
}
```

@NGTmeaty Because you wrote it, could you add a test that verifies the above code is functioning as intended?

### DIFF
--- a/internal/pkg/utils/url_string.go
+++ b/internal/pkg/utils/url_string.go
@@ -25,6 +25,7 @@ func URLToString(u *url.URL) string {
 		tempHost = u.Hostname()
 	}
 
+	// Handle IPv6 Address Formatting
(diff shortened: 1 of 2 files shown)


willmhowes:
I understand now that the code I referenced above is to handle IPv6 addresses, so I added a test for that functionality as well as a comment in the `URLToString` function to explain that line which I consider not immediately obvious. @NGTmeaty You can disregard my request above, but feel free to add any test cases I'm missing

CorentinB:
Yes indeed, it's for IPV6 IPs to have opening and closing brackets.

PR #98, opened 2024-07-26
Add dequeue enqueue stats

CorentinB:


internal/pkg/queue/stats.go
```diff
@@ -174,9 +174,6 @@ func (q *PersistentGroupedQueue) calculateAverageDuration(samples []Sample) OpAv
 }
 
 func (q *PersistentGroupedQueue) addSample(duration time.Duration, sampleType SampleType) {
-	q.stats.Lock()
```
equals215: you sure this will not cause any data race?

equals215:
also can we display the stats using `--live-stats` flag?

CorentinB:
> also can we display the stats using `--live-stats` flag?

Nah, I don't think we should add too much stuff to live-stats.. We already added a lot with the recent PRs.

internal/pkg/queue/stats.go
CorentinB: > you sure this will not cause any data race?

I'm not sure any lock is needed at the stats level here considering the entire queue struct (which hold the stats) is locked in Enqueue and Dequeue ?

PR #104, opened 2024-07-31
Optimize `get list` loading performance

yzqzss:
The time to load 10,000,000 URLs was reduced from 70s to 15s on my machine.

Printing progress for every line is expensive.


CorentinB (approved):
LGTM! Thanks!

PR #105, opened 2024-08-01
Remove `runtime.Gosched()` in polling

yzqzss:
revert #101

Do not use `runtime.Gosched()` in loop, which will spinning your CPUs time for endless coroutine context switching. It won't magically turn polling into blocking.

In addition, Golang's coroutines are preemptive since Go1.14. There will be no starving coroutines, we don't need to manually call `runtime.Gosched()` nowadays.

### DIFF
--- a/internal/pkg/crawl/worker.go
+++ b/internal/pkg/crawl/worker.go
@@ -133,7 +133,7 @@ func (w *Worker) Run() {
 // unsafeCapture is named like so because it should only be called when the worker is locked
 func (w *Worker) unsafeCapture(item *queue.Item) {
 	if item == nil {
-		panic("Item is nil")
+		return
 	}
 
 	// Signals that the worker is processing an item
--- a/internal/pkg/crawl/worker.go
+++ b/internal/pkg/crawl/worker.go
@@ -1,7 +1,6 @@
 package crawl
 
 import (
-	"runtime"
 	"sync"
 	"time"
 
@@ -78,7 +77,7 @@ func (w *Worker) Run() {
 				w.state.lastAction = "waiting for queue to be filled"
 			}
 			if (w.pool.Crawl.Paused.Get() || w.pool.Crawl.Queue.Empty.Get()) && w.pool.Crawl.Queue.CanDequeue() {
-				runtime.Gosched()
+				time.Sleep(10 * time.Millisecond)
 				continue
 			}
 		}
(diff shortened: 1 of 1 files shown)


internal/pkg/crawl/worker.go
```diff
@@ -78,7 +77,7 @@ func (w *Worker) Run() {
 				w.state.lastAction = "waiting for queue to be filled"
 			}
 			if (w.pool.Crawl.Paused.Get() || w.pool.Crawl.Queue.Empty.Get()) && w.pool.Crawl.Queue.CanDequeue() {
```
yzqzss, line 79: Most network requests cannot be completed within 10ms.

Based on this, I think the 10ms polling will not affect the performance of the worker. And if the queue is not empty, `time.Sleep()` will not be executed.

internal/pkg/crawl/worker.go
```diff
@@ -158,7 +157,7 @@ func (w *Worker) unsafeCapture(item *queue.Item) {
 func (w *Worker) Stop() {
 	w.doneSignal <- true
 	for w.state.status != completed {
-		runtime.Gosched()
```
yzqzss, line 161: This led me to observe unusual CPU usage while waiting for workers to shut down.

internal/pkg/crawl/worker.go
```diff
@@ -134,7 +133,7 @@ func (w *Worker) Run() {
 // unsafeCapture is named like so because it should only be called when the worker is locked
 func (w *Worker) unsafeCapture(item *queue.Item) {
 	if item == nil {
-		return
+		panic("Item is nil")
```
CorentinB, line 136: Do we really want to panic? Maybe a log + return would be wiser. (even though this should never happen, right)

btw, error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation in Go

CorentinB (changes requested):
Thank you, I just made a little suggestion. @equals215 will maybe say more as runtime.Gosched was his idea and it was his implementation. Thanks for the contribution once again.

internal/pkg/crawl/worker.go
equals215: I don't understand the reasoning behind the network request, please elaborate

yzqzss: Since we are polling every 10ms here, it will take 10ms for the woker to react to the change of `(w.pool.Crawl.Paused.Get() || w.pool.Crawl.Queue.Empty.Get()) && w.pool.Crawl.Queue.CanDequeue()` from false to true.

I mean 10ms is reasonable.

internal/pkg/crawl/worker.go
yzqzss: oh, I accidentally put the change in this commit :(

Yeah, that will never happen.

equals215:
Yeah revet gosched, lgtm

PR #106, opened 2024-08-01
Queue handover v2

equals215:
Add :
- make handover optional
- added lastAction on API worker state to troubleshoot
- made the default HQ batches smaller
- made handover hold the whole batch before enqueuing on disk
- handover now has an automatic closing mechanism that adapts to activity so the handover get closed and drained on disk if no workers need it


cmd/get.go
```diff
@@ -56,6 +56,7 @@ func getCMDsFlags(getCmd *cobra.Command) {
 	getCmd.PersistentFlags().StringSlice("exclude-string", []string{}, "Discard any (discovered) URLs containing this string.")
 	getCmd.PersistentFlags().Bool("random-local-ip", false, "Use random local IP for requests. (will be ignored if a proxy is set)")
 	getCmd.PersistentFlags().Int("min-space-required", 20, "Minimum space required in GB to continue the crawl.")
+	getCmd.PersistentFlags().Bool("use-handover", false, "Use handover mechanism to dispatch URLs via a buffer before enqueuing on disk.")
```
CorentinB, line 59: I'd use "handover" instead of "use-handover"

CorentinB (changes requested):

PR #107, opened 2024-08-02
Improve WAL concurrency performance

yzqzss:
~~WIP~~ DONE

I'm a bit sleepy, I'll add the PR description hours later.

### DIFF
--- a/internal/pkg/crawl/crawl.go
+++ b/internal/pkg/crawl/crawl.go
@@ -164,13 +164,27 @@ func (c *Crawl) Start() (err error) {
 		// Push the seed list to the queue
 		c.Log.Info("Pushing seeds in the local queue..")
 		var seedPointers []*queue.Item
-		for _, item := range c.SeedList {
+		for idx, item := range c.SeedList {
 			seedPointers = append(seedPointers, &item)
+
+			// We enqueue seeds by batch of 100k
+			// Workers will start processing them as soon as one batch is enqueued
+			if idx%100000 == 0 {
+				c.Log.Info("Enqueuing seeds", "index", idx)
+				if err := c.Queue.BatchEnqueueUntilCommited(seedPointers...); err != nil {
+					c.Log.Error("unable to enqueue seeds, discarding", "error", err)
+				}
+				seedPointers = nil
+			}
 		}
-		c.Queue.BatchEnqueueUntilCommited(seedPointers...) // TODO: chunk
+		if len(seedPointers) > 0 {
+			if err := c.Queue.BatchEnqueueUntilCommited(seedPointers...); err != nil {
+				c.Log.Error("unable to enqueue seeds, discarding", "error", err)
+			}
+		}
+
 		c.SeedList = nil
-		seedPointers = nil
-		c.Log.Info("All seeds are now in queue, crawling will start")
+		c.Log.Info("All seeds are now in queue")
 	}
 
 	// Start the background process that will catch when there
--- a/internal/pkg/crawl/crawl.go
+++ b/internal/pkg/crawl/crawl.go
@@ -171,14 +171,14 @@ func (c *Crawl) Start() (err error) {
 			// Workers will start processing them as soon as one batch is enqueued
 			if idx%100000 == 0 {
 				c.Log.Info("Enqueuing seeds", "index", idx)
-				if err := c.Queue.BatchEnqueueUntilCommited(seedPointers...); err != nil {
+				if err := c.Queue.BatchEnqueueUntilCommitted(seedPointers...); err != nil {
 					c.Log.Error("unable to enqueue seeds, discarding", "error", err)
 				}
 				seedPointers = nil
 			}
 		}
 		if len(seedPointers) > 0 {
-			if err := c.Queue.BatchEnqueueUntilCommited(seedPointers...); err != nil {
+			if err := c.Queue.BatchEnqueueUntilCommitted(seedPointers...); err != nil {
 				c.Log.Error("unable to enqueue seeds, discarding", "error", err)
 			}
 		}
--- a/internal/pkg/crawl/crawl.go
+++ b/internal/pkg/crawl/crawl.go
@@ -163,11 +163,13 @@ func (c *Crawl) Start() (err error) {
 	} else {
 		// Push the seed list to the queue
 		c.Log.Info("Pushing seeds in the local queue..")
+		var seedPointers []*queue.Item
 		for _, item := range c.SeedList {
-			item := item
-			c.Queue.Enqueue(&item)
+			seedPointers = append(seedPointers, &item)
 		}
+		c.Queue.BatchEnqueueUntilCommited(seedPointers...) // TODO: chunk
 		c.SeedList = nil
+		seedPointers = nil
 		c.Log.Info("All seeds are now in queue, crawling will start")
 	}
 
--- a/internal/pkg/queue/enqueue.go
+++ b/internal/pkg/queue/enqueue.go
@@ -10,7 +10,11 @@ import (
 // BatchEnqueue adds 1 or many items to the queue in a single operation.
 // If multiple items are provided, the order in which they will be enqueued is not guaranteed.
 // It WILL be less efficient than Enqueue for single items.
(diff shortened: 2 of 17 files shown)


internal/pkg/crawl/crawl.go
```diff
@@ -163,11 +163,13 @@ func (c *Crawl) Start() (err error) {
...
 		c.Log.Info("Pushing seeds in the local queue..")
+		var seedPointers []*queue.Item
 		for _, item := range c.SeedList {
-			item := item
-			c.Queue.Enqueue(&item)
+			seedPointers = append(seedPointers, &item)
 		}
+		c.Queue.BatchEnqueueUntilCommited(seedPointers...) // TODO: chunk
 		c.SeedList = nil
+		seedPointers = nil
```
equals215, line 172: nulling a pointer used in a function without waiting for the function return can be deadly, this is not safe

internal/pkg/queue/enqueue.go
```diff
@@ -160,32 +170,42 @@ func (q *PersistentGroupedQueue) Enqueue(item *Item) error {
...
-	q.mutex.Lock()
-	defer q.mutex.Unlock()
+	var commit uint64 = 0
 
-	// Find free position
-	startPos, err := q.queueFile.Seek(0, io.SeekEnd)
-	if err != nil {
-		return fmt.Errorf("failed to seek to end of file: %w", err)
-	}
+	err := func() error {
```
equals215, line 175: what's the point of using this syntax? you can just return on error of any of the function calls, the mutex will still be deferred

internal/pkg/queue/index/manager.go
```diff
@@ -57,22 +72,29 @@ func NewIndexManager(walPath, indexPath, queueDirPath string) (*IndexManager, er
...
+		WAL_IO_PERCENT:     10,
+		WAL_MIN_INTERVAL:   10 * time.Millisecond,
+		walStopChan:        make(chan struct{}),
+		stopChan:           make(chan struct{}),
 	}
 
 	// Check if WAL file is empty
 	im.Lock()
-	empty, err := im.unsafeIsWALEmpty()
+	empty, err := im.unsafeIsWALEmpty() // FIXME: check error
```
equals215, line 97: spotted FIXME

internal/pkg/queue/index/manager.go
```diff
@@ -29,23 +30,37 @@ type WALEntry struct {
...
+	walDecoder  *gob.Decoder
+	walCommit   *atomic.Uint64 // Flying in memory commit id
+	walCommited *atomic.Uint64 // Synced to disk commit id
+	// Number of listeners waiting for walCommitedNotify.
+	// It must be accurate, otherwise walNotifyListeners will be blocked
+	walNotifyListeners *atomic.Int64
+	walCommitedNotify  chan uint64 // receives the commited id from walCommitsSyncer
+	walSyncerRunning   atomic.Bool // used to prevent multiple walCommitsSyncer running,
+	walStopChan        chan struct{}
+	WAL_IO_PERCENT     int           // [1, 100] limit max io percentage for WAL sync
```
equals215, line 55: all caps + underscore var name doesn't respect go style/idiom

internal/pkg/queue/index/manager.go
```diff
@@ -108,39 +131,130 @@ func NewIndexManager(walPath, indexPath, queueDirPath string) (*IndexManager, er
...
+		// fmt.Println("lastSyncDuration", lastSyncDuration, "sleepTime", sleepTime)
+		time.Sleep(sleepTime)
+
+		start := time.Now()
+		flyingCommit := im.walCommit.Load()
+		im.Lock()
+		if err := im.unsafeWalSync(); err != nil {
+			slog.Error("failed to sync WAL", "error", err)
+			im.Unlock()
+			return // FIXME: handle this error
```
equals215, line 168: spotted FIXME

internal/pkg/queue/index/manager.go
```diff
@@ -187,6 +305,10 @@ func (im *IndexManager) Pop(host string) (id string, position uint64, size uint6
 func (im *IndexManager) Close() error {
 	im.dumpTicker.Stop()
 	im.stopChan <- struct{}{}
+	im.walStopChan <- struct{}{}
+	if im.walSyncerRunning.Load() {
+		panic("walCommitsSyncer still running")
```
equals215, line 310: panic because a goroutine is still running is not an acceptable way of handling. this will make the other components fail to close gracefully for a simple error

equals215 (changes requested):
That's a cool feature, no doubt!
As @CorentinB already told you it hangs when used with our crawl scheduler (HQ) which is closed source. So on our end we can't really accept that on our repo as-is.

Also having a committed WAL that is synced on disk when full defeats the purpose of why we implemented an atomic WAL in the first place : we're trying to focus reliability of the queue so that the least seeds possible are lost in the event of a crash.

We need to discuss that further, if you're willing to make this feature optional with a flag we will be more inclined to accept your contribution!

internal/pkg/crawl/crawl.go
```diff
@@ -163,11 +163,13 @@ func (c *Crawl) Start() (err error) {
 	} else {
 		// Push the seed list to the queue
 		c.Log.Info("Pushing seeds in the local queue..")
+		var seedPointers []*queue.Item
 		for _, item := range c.SeedList {
-			item := item
-			c.Queue.Enqueue(&item)
+			seedPointers = append(seedPointers, &item)
 		}
+		c.Queue.BatchEnqueueUntilCommited(seedPointers...) // TODO: chunk
```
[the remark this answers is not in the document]
yzqzss: On my machine, `BatchEnqueueUntilCommited()` took 10～20 seconds for 10 million items. (resulted in 2.6G queue file)

No one here is using `get list` to load hundreds of millions of URLs, right?
If so, it would be better to call `BatchEnqueueUntilCommited()` once for every few million URLs.
:)

internal/pkg/crawl/crawl.go
yzqzss: `seedPointers` is just used to convert c.SeedList (`[]queue.Item`) to `[]*queue.Item` .
Anyway, `seedPointers = nil` is useless here, I deleted the line.

internal/pkg/queue/enqueue.go
yzqzss: image

I want to ensure that the execution of this closure (yellow) is locked, but the subsequent green area does not need to be locked.

internal/pkg/queue/index/manager.go
yzqzss: https://github.com/internetarchive/Zeno/commit/7c6b1066729882a1c2a344cb4692c9962f8efe8f

I'm not sure about the intention here.

internal/pkg/crawl/crawl.go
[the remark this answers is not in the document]
yzqzss: done via 1c80c199e6fa98cdb581a15dc470cc6ab4ea602e

internal/pkg/queue/index/manager.go
yzqzss: renamed

internal/pkg/queue/index/manager.go
yzqzss: 04e02e60e0002137d6ec0bcf3e81cb081f9778c6

Now it is changed to retry if fsync fails. it's better than losing data.

internal/pkg/queue/index/manager.go
yzqzss: 04e02e60e0002137d6ec0bcf3e81cb081f9778c6

walCommitsSyncer() now shuts down gracefully

internal/pkg/queue/index/manager.go
equals215: you're right the error is not checked after return

internal/pkg/queue/enqueue.go
equals215: Ok got it, you might wanna declare a function outside the scope of EnqueueUntilCommitted() for clarity of code and debugging. Unnamed functions/routines are not recommended.

equals215:
Merged so I can work on queue while you sleep tight @yzqzss

PR #117, opened 2024-08-05
Enable `log` package to distribute a stored logger to all other packages

equals215:


CorentinB (approved):
LGTM (but test doesn't pass)

equals215:
> Looks good to me (but test doesn't pass)

weird I might have broken smth, on it

equals215:
found the error : `once.Do()` is called inside of a `once.Do()` when `DefaultOrStored()` is called and no logger exists already

equals215:
tests wont pass due to #118, merging

PR #120, opened 2024-08-06
Calming down queue stats

yzqzss:
- remove HostDistribution and TopHosts stats

`genStats()` takes too much time to generate HostDistribution and TopHosts.

Let's remove them until we find a better statistics implementation. :disappointed:


CorentinB (approved):
LGTM

PR #122, opened 2024-08-06
[STALE] Split Zeno in smaller packages with a better structure

equals215:


equals215:
## Notes :

### Init sequence :
1. Log
2. Stats
3. Queue

equals215:
TODO : Make processor workerPool/queue agnostic to it can be used with pure channels and redistribute outlinks to other workers

equals215:
1:30am notes 
Communication between reactor/bus is not working and/or not properly done resulting in wasting seeds
Workers don't get their seeds due to the select case defaulting?
HQ is perma de queuing-> use stats to get queue size?

PR #126, opened 2024-08-12
Add proper YouTube archiving via YT-DLP

CorentinB:


cmd/get.go
```diff
@@ -84,6 +84,10 @@ func getCMDsFlags(getCmd *cobra.Command) {
 	getCmd.PersistentFlags().String("es-password", "", "ElasticSearch password to use for indexing crawl logs.")
 	getCmd.PersistentFlags().String("es-index-prefix", "zeno", "ElasticSearch index prefix to use for indexing crawl logs. Default is : `zeno`, without `-`")
 
+	// Dependencies flags
+	getCmd.PersistentFlags().Bool("no-ytdlp", false, "Disable youtube-dlp usage for video extraction.")
```
equals215, line 88: maybe: rename the flag to `--use-ytdlp` and make the default behaviour not use `yt-dlp` as you don't want external dependencies by default

internal/pkg/crawl/crawl.go
```diff
@@ -125,6 +126,20 @@ func (c *Crawl) Start() (err error) {
 		go c.startAPI()
 	}
 
+	// Verify that dependencies exist on the system
+	if !c.NoYTDLP {
```
equals215, line 130: instead of adding (another) code block to `crawl.Start()` maybe make a function in `dependencies` package that check if dependencies are satisfied when enabled for the crawl

internal/pkg/crawl/dependencies/ytdlp/model.go
equals215, line 1: rename `dependencies` package to `external` or `external/dependencies` in opposition to `internal`

internal/pkg/crawl/dependencies/ytdlp/parse.go
```diff
@@ -0,0 +1,42 @@
...
+type SubtitleInfo struct {
+	Ext  string `json:"ext"`
+	URL  string `json:"url"`
+	Name string `json:"name"`
+}
+
+// parseSubtitles parses the subtitles from the yt-dlp JSON output,
+// it's needed because the subtitles are not given as a proper array or objects
+func parseSubtitles(jsonData string) ([]string, error) {
+	var data map[string]json.RawMessage
```
equals215, line 17: init the map with `make()`

internal/pkg/crawl/sitespecific/youtube/youtube.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/internetarchive/Zeno/internal/pkg/crawl/dependencies/ytdlp"
+)
+
+func Parse(body io.ReadCloser) (URLs []*url.URL, err error) {
+	// Create a temporary server to serve the body and call ytdlp on it
+	port, stopChan, err := ytdlp.ServeBody(body)
+	if err != nil {
+		return nil, err
+	}
+	defer close(stopChan)
```
equals215, line 16: maybe close the channel in `internal/pkg/crawl/dependencies/ytdlp/server.go#L32` and just send a `struct{}{}` here, it feels more logical to me

equals215 (changes requested):
nice job, it doesn't even feel hacky as you said it was

cmd/get.go
CorentinB: No I think we want to use yt-dlp by default, but if it's not on the system then it won't be used, see what I mean?

equals215: I see

PR #129, opened 2024-08-13
Add additional tests and validate current URL behavior

NGTmeaty:
This PR adds additional tests and removes unneeded calls when turning URLs into strings.


CorentinB (approved):
LGTM

PR #140, opened 2024-08-22
Fix worker state panic when current item or URL is nil

### LINKED ISSUE #132: Panic on /workers access
```
2024/08/15 07:49:36 http: panic serving 127.0.0.1:45290: runtime error: invalid memory address or nil pointer dereference
goroutine 212832422 [running]:
net/http.(*conn).serve.func1()
        /var/www/.go/src/net/http/server.go:1903 +0xbe
panic({0x1371ac0?, 0x21a7f70?})
```

CorentinB:
Fixes #132


equals215 (commented):
LGTM

PR #142, opened 2024-08-29
fix: `HQSeencheckURL()` incorrectly distinguishes between new and old URL

yzqzss:
the returned payload from HQ only contain new URL(s) to be crawled by Zeno.


NGTmeaty (approved):
LGTM! Thank you!

PR #170, opened 2024-11-25
Add pipeline control mechanism to instantiate, pause/resume and stop

equals215:

PR #171, opened 2024-11-25
Add log config support

equals215:


equals215:
@CorentinB ready to test and merge if successful

equals215:
@CorentinB ready to be tested and merged if satisfying

internal/pkg/log/config.go
```diff
@@ -2,45 +2,96 @@
...
+	Level        slog.Level
+	Rotate       bool
+	RotatePeriod time.Duration
 }
 
-// defaultConfig returns the default configuration
-func defaultConfig() *Config {
+// makeConfig returns the default configuration
+func makeConfig() *Config {
+	fileRotatePeriod, err := time.ParseDuration(config.Get().LogFileRotation)
```
CorentinB, line 45: I get a panic at this line when calling Zeno (get url..) with `-h`:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x298 pc=0xdf38e1]

goroutine 1 [running]:
github.com/internetarchive/Zeno/internal/pkg/log.makeConfig()
	/home/corentin/Documents/work/ia/Zeno/internal/pkg/log/config.go:45 +0x21
github.com/internetarchive/Zeno/internal/pkg/log.Start.func1()
	/home/corentin/Documents/work/ia/Zeno/internal/pkg/log/log.go:26 +0x1b
sync.(*Once).doSlow(0x116904c?, 0x412985?)
	/home/corentin/.go/src/sync/once.go:76 +0xb4
sync.(*Once).Do(...)
	/home/corentin/.go/src/sync/once.go:67
github.com/internetarchive/Zeno/internal/pkg/log.Start()
	/home/corentin/Documents/work/ia/Zeno/internal/pkg/log/log.go:25 +0x45
main.main()
	/home/corentin/Documents/work/ia/Zeno/main.go:40 +0x3f
```

CorentinB (changes requested):
I wish there would be an easy way for Zeno to log to its job dir. Considering that logging is not enabled by default, and that job path is made out of a random string, one cannot log to a `logs` dir in the jobs dir.

internal/pkg/log/config.go
equals215: Oh that's cause we run the pipeline from main and not from a cmd func (which is a bad practice)
So when you pass `-h`, `spf13/cobra`'s helper func runs and then the rest of the main is executed _BUT_ the config haven't been initialized and the call to `config.Get().<whatever>` segfaults cause the `config == nil`
We need to init the pipeline at the `cmd` level.

equals215: Resolved by #170 and last commit

PR #173, opened 2024-11-29
Fix various bugs in the pipeline

CorentinB:


equals215:
Why is this guy so bad with git

CorentinB:
> Why is this guy so bad with git

not nice, not nice.....

PR #185, opened 2025-01-24
Add reddit support to v2

NGTmeaty:


internal/pkg/postprocessor/extractor/html.go
```diff
@@ -361,6 +363,12 @@ func HTMLAssets(item *models.Item) (assets []*models.URL, err error) {
 	}
 
 	for _, rawAsset := range rawAssets {
+		if reddit.IsRedditURL(item.GetURL()) {
```
CorentinB, line 366: Can you instead do that after the HTML extraction, in the postprocessor, so that all site-specific activations take place in the same function? Same goes for line 31 of json.go.

CorentinB (changes requested):

internal/pkg/postprocessor/extractor/html.go
NGTmeaty: It was originally only in those specific extractors because those were the only ones that needed it, but I agree that it is more elegant to put it in the postprocessor.

PR #189, opened 2025-01-28
Better JSON/XML extraction

CorentinB:
This PR adds a check when doing JSON & XML extraction of assets to verify that the URLs we extract are indeed files and not "pages". URLs that do not pass the test are returned as outlinks to be (potentially) queued back.


equals215 (approved):
LGTM

PR #191, opened 2025-01-30
Reduce stack usage

equals215:
We are observing solid discrepancies between the heap usage as profiled by Pyroscope and what the system reports as used. As Pyroscope only profiles the heap (mem allocated by go runtime) it suggests that the stack is being used too much and is causing the OOMs observed. Most likely causes of an off-heap usage are :
- Goroutine Stacks
- CGO allocations
- Memory-mapped files
- Go Runtime Overhead

We are only using one CGO library : `goada` and a change was rolled to forcefully free any object allocated via this library (on the stack as the C++ code calls `malloc()`)

Second line of investigation would be to troubleshoot WARC library as it's the top producer of goroutines and the profiler shows a lot of blocking : maybe the routines get stuck indefinitely and increase stack usage


CorentinB (approved):

equals215:
Will not merge, splitting the PR in multiple PR with the change we want to keep

PR #201, opened 2025-02-03
Fix outlinks extraction

CorentinB:


internal/pkg/postprocessor/extractor/resolve.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// If the link is already absolute, return it.
+	if link.IsAbs() {
+		return link.String(), nil
+	}
+
+	// Resolve the relative URL against the base.
+	// The net/url.ResolveReference method follows RFC 3986, handling
+	// relative paths (including those starting with "/" or "../").
+	return baseURL.ResolveReference(link).String(), nil
```
equals215, line 41: I don't really like that due to infosec concerns...

equals215 (approved):
LGTM

PR #209, opened 2025-02-07
Rework concurrency for pipeline stages

equals215:
This PR itself serves as a point of centralization for every concurrency rework PR. This shouldn't be reviewed, upstream PR should be.
Every stage is reworked to use the classic fan-in/fan-out concurrency model instead of the dynamic one we made at the start which might be causing deadlocks and fatigue on the system.

### DIFF
--- a/internal/pkg/archiver/archiver.go
+++ b/internal/pkg/archiver/archiver.go
@@ -43,7 +43,7 @@ var (
 	logger         *log.FieldedLogger
 )
 
-// This functions starts the archiver responsible for capturing the URLs
+// Start initializes the internal archiver structure, start the WARC writer and start routines, should only be called once and returns an error if called more than once
 func Start(inputChan, outputChan chan *models.Item) error {
 	var done bool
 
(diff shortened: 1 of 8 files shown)


internal/pkg/preprocessor/preprocessor.go
```diff
@@ -76,204 +88,192 @@ func Stop() {
 	}
 }
 
-func run() {
+func (p *preprocessor) worker(workerID string) {
```
equals215, line 91: `p *preprocessor` is not used

internal/pkg/preprocessor/preprocessor.go
```diff
@@ -76,204 +88,192 @@ func Stop() {
...
+	defer globalPreprocessor.wg.Done()
+
 	logger := log.NewFieldedLogger(&log.Fields{
 		"component": "preprocessor.run",
```
equals215, line 95: should be "preprocessor.worker"

internal/pkg/archiver/archiver.go
```diff
@@ -113,75 +119,62 @@ func Stop() {
...
+
+				if seed.GetStatus() != models.ItemPreProcessed && seed.GetStatus() != models.ItemGotRedirected && seed.GetStatus() != models.ItemGotChildren {
+					logger.Debug("skipping seed", "seed", seed.GetShortID(), "depth", seed.GetDepth(), "hops", seed.GetURL().GetHops(), "status", seed.GetStatus().String())
+				} else {
+					archive(workerID, seed)
+				}
+
+				select {
+				case a.outputCh <- seed:
+				case <-a.ctx.Done():
```
equals215, line 164: `ctx.Done()` check should be before passing the seed to the next stage

CorentinB (approved):

PR #216, opened 2025-02-09
Optimise extractBaseTag and add unit test

vbanos:
The current code looks for `base` tag but doesn't stop if it finds one. It will still search until the end of the doc.

The suggested improvement uses `doc.Find("base").First()` to get just the first element as `<base>` is used just once on the HTML doc header.

We also add a unit test.


CorentinB:
LGTM! Thanks

PR #220, opened 2025-02-11
Improve isStatusCodeRedirect

vbanos:
Add missing status 303 "See Other"
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303

Refactor to make it more readable.


CorentinB:
How did I miss 303! Thanks.

Issue #221, opened 2025-02-11
Drop redundant Item.seed


vbanos:
`Item.seed` is redundant because  we already have the `parent` attribute. (When `parent == nil`, then the item is a seed).
We must refactor and drop it.

equals215:
Done in #222

PR #222, opened 2025-02-11
Drop item.seed attribute

vbanos:
Item.seed is redundant because we already have the parent attribute. (When parent == nil, then the item is a seed).

Fixing issue: https://github.com/internetarchive/Zeno/issues/221


CorentinB (approved):
LGTM, @equals215 can you also review? Thanks

equals215:
Tests are passing, seems ok to me! Merging

PR #225, opened 2025-02-13
Rework log to use `samber/slog-multi` as a `log/slog` routing abstraction

equals215:
The log package was initially designed to be asynchronous, causing too much log queue full errors and allocating too many objects on the heap.
The rewrite aims to make this package provide the same features but with a robust and well tested slog router, avoiding all the aforementioned issued from the original and poorly implemented logging package.


equals215:
Elasticsearch logging not reimplemented cause I'm pretty sure logging via the API is not the best method possible and I can do better use of my time. Ready to merge.

CorentinB (approved):

PR #227, opened 2025-02-16
Refactor models.URL goquery document parsing

vbanos:
Remove goquery from the URL model and put it in the HTML extractor.

### DIFF
--- a/internal/pkg/postprocessor/extractor/html.go
+++ b/internal/pkg/postprocessor/extractor/html.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/CorentinB/warc/pkg/spooledtempfile"
 	"github.com/PuerkitoBio/goquery"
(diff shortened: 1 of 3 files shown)


vbanos:
We have the URL model which is generic and it has these methods to parse HTML and create a document using goquery.
URL could be any mime type but we don't have methods for any other mime type besides HTML.
Also, these methods are only used in internal/pkg/postprocessor/html.go.
Therefore, I suggest to remove any goquery related code from the model and put it in internal/pkg/postprocessor/html.go where it belongs.

vbanos:
One performance observation regarding this PR:

With the earlier code `URL.document` is created when you run `HTMLAssets` and then its reused when you run `HTMLOutlinks`.

With this PR, it needs to run again.

IMHO this issue should be handled internally in `internal/pkg/postprocessor/extractor/html.go`.
If we add an attribute to the model for each extractor type, the model with become a mess.

PR #233, opened 2025-02-24
Expose Prometheus metrics and make Zeno self-declare as a Consul service for it's metrics to be collected automatically

equals215:
- [x] Expose Prometheus endpoint
- [x] Collect `stats` data and expose them via the Prom endpoint
- [x] Integrate Consul service self-declaration


CorentinB:
Would be interesting to have counters for HTTP errors.

CorentinB:
RAM available and used too maybe?

equals215:
> "Would be interesting to have counters for HTTP errors."
> "RAM available and used too maybe?"

Will be done in a new PR to avoid overloading this one with features

internal/pkg/consul/consul.go
```diff
@@ -0,0 +1,96 @@
...
+			DeregisterCriticalServiceAfter: "30s",
+		},
+	}
+
+	// Register the service with Consul.
+	if err := client.Agent().ServiceRegister(registration); err != nil {
+		return fmt.Errorf("failed to register service: %v", err)
+	}
+
+	logger.Debug("registered service", "serviceID", serviceID, "serviceName", serviceName, "serviceTags", serviceTags)
```
CorentinB, line 70: Might be worth be put in Info, it's not like it's printed 1000 times.

internal/pkg/consul/consul.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Start a goroutine to periodically update the TTL check.
+	go func() {
+		ticker := time.NewTicker(10 * time.Second)
+		defer ticker.Stop()
+		for {
+			// Update the TTL to indicate the service is healthy.
+			err := client.Agent().UpdateTTL("service:"+serviceID, "passing", api.HealthPassing)
+			if err != nil {
+				log.Error("failed to update TTL", "serviceID", serviceID, "error", err)
```
CorentinB, line 80: Any reason it's not using our usual logger?

internal/pkg/consul/consul.go
```diff
@@ -0,0 +1,96 @@
...
+			}
+
+			select {
+			case <-ticker.C:
+			case <-ctx.Done():
+				// De-register the service when the context is canceled.
+				if err := client.Agent().ServiceDeregister(serviceID); err != nil {
+					log.Error("failed to deregister service", "serviceID", serviceID, "error", err)
```
CorentinB, line 88: Any reason it's not using our usual logger?

CorentinB (changes requested):

PR #248, opened 2025-03-14
Refactor: replace timers with tickers for batch processing

yzqzss:
Just a refactor, nothing should be affected.


CorentinB:
Thanks!

PR #249, opened 2025-03-17
Extract RSS XML as general XML

yzqzss:
XML RSS is just XML; the only thing preventing Zeno post-processing for XML RSS is its mimetype.Parent().String() != "text/plain".

https://github.com/internetarchive/Zeno/blob/f004ca45ea0a06d6958d21be919f5b8ca34ddb80/internal/pkg/archiver/body.go#L46-L47

General XML mimetype:
"text/xml; charset=utf-8" -> "text/plain" -> "application/octet-stream"
`u.GetMIMEType().Parent().String() == "text/plain"` is `true`

XML RSS mimetype:
"application/rss+xml" -> "text/xml" -> "text/plain" -> "application/octet-stream"
`u.GetMIMEType().Parent().String() == "text/plain"` is `false`

---

To address this, I added `IsMIMETypeInHierarchy()` to recursively check if the MIME type and its parents match the expected MIME type.

close: #159

### DIFF
--- a/internal/pkg/archiver/body.go
+++ b/internal/pkg/archiver/body.go
@@ -9,6 +9,7 @@ import (
 	"github.com/CorentinB/warc/pkg/spooledtempfile"
 	"github.com/gabriel-vasile/mimetype"
 	"github.com/internetarchive/Zeno/internal/pkg/config"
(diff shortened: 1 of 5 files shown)


CorentinB:
Thanks!

PR #251, opened 2025-03-19
PDF extractor

yzqzss:
Extract PDF links by using https://github.com/pdfcpu/pdfcpu

For now, we only extract "clickable" links (`LinkAnnotation`) from PDF.
Plain text url extraction is possible, but I'm afraid it won't work well.

---

close: #161


CorentinB:
LGTM! Thanks.

PR #256, opened 2025-03-28
Integrate rate limiting in archiver and implement retries

yzqzss:


cmd/get.go
```diff
@@ -63,6 +65,12 @@ func getCMDsFlags(getCmd *cobra.Command) {
 	getCmd.PersistentFlags().Bool("disable-ipv6", false, "Disable IPv6 for requests.")
 	getCmd.PersistentFlags().Bool("ipv6-anyip", false, "Use AnyIP kernel feature for requests. (only IPv6, need --random-local-ip)")
 
+	// Rate limiting flags
+	getCmd.PersistentFlags().Bool("disable-rate-limit", false, "Disable the Token Bucket rate limiting.")
+	getCmd.PersistentFlags().Float64("rate-limit-capacity", 10, "Bucket capacity for each host.")
+	getCmd.PersistentFlags().Float64("rate-limit-refill-rate", 2, "Ideal requests per second for each host.")
```
yzqzss, line 71: I have no idea on chosing a sweet --rate-limit-refill-rate default value.

internal/pkg/archiver/archiver.go
```diff
@@ -248,6 +288,14 @@ func archive(workerID string, seed *models.Item) {
 
 			stats.HTTPReturnCodesIncr(strconv.Itoa(resp.StatusCode))
 
+			if globalBucketManager != nil {
+				if resp.StatusCode == 200 {
```
yzqzss, line 292: should we check other 2XX code?

cmd/get.go
CorentinB: I think 50 is a good default. Most hosts won't mind the rate and when there is a problem it will reduce!

internal/pkg/archiver/archiver.go
CorentinB: Yes definitely, also technically a success can be like 404 or other like that.

CorentinB: I think a failure should be 429, 5XX of course, possibly 403 but it's trickier.

CorentinB: Moving to the feat/retries branch.

PR #252, opened 2025-03-19
EPUB Support

akshithio:
Hi, I'm Akshith, and I'm interested in contributing to Zeno for GSoC 2025. This PR is my attempt at familiarizing myself with the codebase — I apologize in advance for any errors.

This PR adds initial support for `epub` files alongside a custom-made `my_epub.epub` in testdata for which test cases are ~~still a work in progress~~ also written and passing. 

Since `epub` files are essentially zipped folders containing `xhtml` and `html` files, `epub.go` currently works by unzipping the file and looking for these assets. I've tried to take inspiration from existing implementations and patterns in `html.go` where I felt applicable. Please advise on whether this is a suitable approach.

Attempts to close #160.

PR #260, opened 2025-04-07
EPUB Support

### LINKED ISSUE #160: Ebook extractor
Need to handle many/all ebook formats.

akshithio:
Copied over from PR #252:

This PR adds initial support for `epub` files alongside a custom-made `my_epub.epub` in testdata for which test cases are ~still a work in progress~ also written and passing.

Since `epub` files are essentially zipped folders containing `xhtml` and `html` files, `epub.go` currently works by unzipping the file and looking for these assets. I've tried to take inspiration from existing implementations and patterns in `html.go` where I felt applicable. Please advise on whether this is a suitable approach.

------

Additionally, I have addressed the issue with the `ensureBasePath` and fixed that to make sure that the implementation is more OS-agnostic.

Attempts to close #160.


akshithio:
@CorentinB

Apologies for the delay but I did make some changes to try to address this issue. I'm not necessarily sure if all of them are more suitable which is why I'd like to hear what you think about them. Since firstly to deal with the folder structure with which relative file paths are mentioned in `.epub` files, I thought it would be best to just mostly re-write that logic with little changes. 

However, I've now made changes to that approach with a shared helper for both outlinks and assets but that means I have also had to change `html.go` and some other files. I have listed those changes in detail here:

- `html_extractor.go` - new file created containing methods to extract outlinks and assets in `html.go` and `epub.go`
- `html.go` - Changed to call the new methods in `html_extractor.go` instead
- `epub.go` -  Changed to call the new methods in `html_extractor.go` instead
- `epub_test.go` - Add two new test links in a previous test case because I was not previously detecting a `.js` file and some other file (I noticed this when re-writing and was accordingly able to use the new `html_extractor.go` file effectively)
- `html_test.go` - Changed to check the returned links instead of the count both make stronger test cases and to help in effective debugging.
- `json.go` - Minor changes to `findURLs` to ensure relative paths are also passed back, which is required for `.epub` asset detection.

CorentinB:
No activity, closing.

PR #263, opened 2025-04-09
Switch to `DiscardHook`

yzqzss:
Pre: https://github.com/CorentinB/warc/pull/91


CorentinB:
Sorry it's not very important but can you switch the discarding hook to a separate package or at least separate file? I tend to think we will quickly expand it into a big pile of code to handle a lot of scenario so better be ready :)

yzqzss:
> Sorry it's not very important but can you switch the discarding hook to a separate package or at least separate file? I tend to think we will quickly expand it into a big pile of code to handle a lot of scenario so better be ready :)

OK.

And to better handle a lot of scenarios, I want to change the hook function to return an additional reason string for better observability. :)

like:
```go
	discardHook := func(resp *http.Response) (bool, string) {
		// Cloudflare challenge
		if resp.StatusCode == 403 && resp.Header.Get("cf-mitigated") == "challenge" {
			return true, "Detected Cloudflare challenge page"
		}

		// --warc-discard-status
		if len(config.Get().WARCDiscardStatus) > 0 && slices.Contains(config.Get().WARCDiscardStatus, resp.StatusCode) {
			return true, "Response status code in --warc-discard-status"
		}

		return false, "No need to discard"
	}
```

internal/pkg/archiver/archiver.go
```diff
@@ -269,15 +271,31 @@ func archive(workerID string, seed *models.Item) {
...
 					}
+
+					retryReason := "bad response code"
+					if isDiscardedChallengePage {
+						retryReason = discardReason
+					}
+
 					if retry < config.Get().MaxRetry {
-						logger.Warn("bad response code, retrying", "seed_id", seed.GetShortID(), "item_id", item.GetShortID(), "depth", item.GetDepth(), "hops", item.GetURL().GetHops(), "retry", retry, "sleep_time", retrySleepTime.String(), "status_code", resp.StatusCode, "url", req.URL.String())
+						logger.Warn("retrying", "why_retry", retryReason, "seed_id", seed.GetShortID(), "item_id", item.GetShortID(), "depth", item.GetDepth(), "hops", item.GetURL().GetHops(), "retry", retry, "sleep_time", retrySleepTime.String(), "status_code", resp.StatusCode, "url", req.URL.String())
```
CorentinB, line 298: I'd suggest `reason` instead of `why_retry`. Or `retry_reason`, as you wish.

yzqzss:
> Go package name shouldn't contain underscores unless really necessary

Ahaha, you're right.
My brain must have failed when I switched naming styles between Python and Go. 😅

PR #272, opened 2025-04-27
Enhance base URL handling

yzqzss:
- Enhance `extractBaseTag()` function to conform with whatwg's spec: <https://html.spec.whatwg.org/multipage/semantics.html#the-base-element>
- Change base URL type in Item model from `string` to `*url.URL` for convenience.
- Make `HTMLAssets()` resolve URLs directly in `html.go`, just like `HTMLOutlinks()`.


CorentinB:
LGTM! Thanks

PR #279, opened 2025-05-12
feat: Extract WACZ files from replayweb embeds

NGTmeaty:


internal/pkg/postprocessor/extractor/html.go
```diff
@@ -377,6 +377,16 @@ func HTMLAssets(item *models.Item) (assets []*models.URL, err error) {
 		})
 	}
 
+	// Extract WACZ files from replayweb embeds
```
willmhowes, line 380: Link `https://replayweb.page/docs/embedding/` in the comments?

willmhowes (approved):
LGTM

PR #281, opened 2025-05-15
Optimize the OSS extraction process and implement Azure Blob extractor

yzqzss:
When fixing #280, I found that Azure Blob is not S3 compatible, so I implemented a separate Azure sub-extractor. Then I thought that one URL can obviously only be one type of OSS, so I created the `ObjectStorage()` method to call the sub-extractor (S3/Azure/...) conditionally.

### DIFF
--- a/internal/pkg/postprocessor/extractor/object_storage.go
+++ b/internal/pkg/postprocessor/extractor/object_storage.go
@@ -1,6 +1,7 @@
 package extractor
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/internetarchive/Zeno/internal/pkg/utils"
@@ -30,6 +31,6 @@ func ObjectStorage(URL *models.URL) ([]*models.URL, error) {
 	} else if utils.StringContainsSliceElements(server, azureServers) {
 		return azure(URL)
 	} else {
-		return nil, nil
+		return nil, fmt.Errorf("unknown object storage server: %s", server)
 	}
 }
--- a/internal/pkg/postprocessor/extractor/object_storage.go
+++ b/internal/pkg/postprocessor/extractor/object_storage.go
@@ -1,0 +1,35 @@
+package extractor
+
+import (
+	"strings"
+
+	"github.com/internetarchive/Zeno/internal/pkg/utils"
+	"github.com/internetarchive/Zeno/pkg/models"
+)
+
+// All the supported object storage servers
+var ObjectStorageServers = func() (s []string) {
+	s = append(s, s3CompatibleServers...)
+	s = append(s, azureServers...)
+	return s
+}()
+
+// IsObjectStorage checks if the response is from an object storage server
+func IsObjectStorage(URL *models.URL) bool {
+	return utils.StringContainsSliceElements(URL.GetResponse().Header.Get("Server"), ObjectStorageServers) &&
+		strings.Contains(URL.GetResponse().Header.Get("Content-Type"), "/xml") // tricky match both application/xml and text/xml
+}
+
+// ObjectStorage decides which helper to call based on the object storage server
+func ObjectStorage(URL *models.URL) ([]*models.URL, error) {
+	defer URL.RewindBody()
+
+	server := URL.GetResponse().Header.Get("Server")
+	if utils.StringContainsSliceElements(server, s3CompatibleServers) {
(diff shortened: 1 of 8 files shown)


internal/pkg/postprocessor/extractor/object_storage.go
```diff
@@ -0,0 +1,35 @@
...
+var ObjectStorageServers = func() (s []string) {
+	s = append(s, s3CompatibleServers...)
+	s = append(s, azureServers...)
+	return s
+}()
+
+// IsObjectStorage checks if the response is from an object storage server
+func IsObjectStorage(URL *models.URL) bool {
+	return utils.StringContainsSliceElements(URL.GetResponse().Header.Get("Server"), ObjectStorageServers) &&
+		strings.Contains(URL.GetResponse().Header.Get("Content-Type"), "/xml") // tricky match both application/xml and text/xml
```
yzqzss, line 20:
```diff
- strings.Contains(URL.GetResponse().Header.Get("Content-Type"), "xml")
+ strings.Contains(URL.GetResponse().Header.Get("Content-Type"), "/xml")
```

Prevent matching `image/svg+xml`, fix #280.

internal/pkg/postprocessor/extractor/object_storage_azure.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Azure Blob Storage
+var azureServers = []string{
+	// Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
+	"Windows-Azure-Blob",
+	// Blob Service Version 1.0 Microsoft-HTTPAPI/2.0
+	"Blob Service Version",
+	// emulator, https://github.com/Azure/Azurite
+	"Azurite-Blob",
+}
```
yzqzss, line 19: I have tested Zeno with the azure extracor locally with [Azurite](https://github.com/Azure/Azurite/) and it works. But I haven't run it on a real Azure Blob yet.

Also, I found that some Azure Blobs use the `Server: Blob Service Version XXX...` header, and I don't know what the difference is between them and `Server: Windows-Azure-Blob...`.

internal/pkg/postprocessor/extractor/object_storage_azure.go
```diff
@@ -0,0 +1,80 @@
...
+	baseURL.RawQuery = ""
+	baseURL.ForceQuery = false
+	if !strings.HasSuffix(baseURL.Path, "/") {
+		baseURL.Path = baseURL.Path + "/"
+	}
+
+	for _, blob := range result.Blobs {
+		fileURL := baseURL
+		if strings.HasPrefix(blob.Name, "/") {
+			panic("TODO")
```
yzqzss, line 73: TODO

internal/pkg/postprocessor/extractor/object_storage_azure.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		fileURL.Path = baseURL.Path + blob.Name
+		outlinks = append(outlinks, fileURL.String())
```
yzqzss, line 76: AZure's `blob.Name` is non-url-encoded path.

image


Concatenate them into `url.URL.Path` so that `url.URL.String()` can encode them correctly.

internal/pkg/postprocessor/extractor/object_storage_azure.go
```diff
@@ -0,0 +1,86 @@
...
+		baseURL.Path = baseURL.Path + "/"
+	}
+
+	for _, blob := range result.Blobs {
+		fileURL := baseURL
+		if strings.HasPrefix(blob.Name, "/") {
+			azureLogger.Warn("invalid blob name: it starts with a leading slash", "blob_name", blob.Name)
+			continue
+		}
+		fileURL.Path = baseURL.Path + blob.Name
```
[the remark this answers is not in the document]
yzqzss: This shouldn't happen unless Microsoft screws up something, I think.

CorentinB (approved):
LGTM! Excellent work, thanks for your contribution! Ready to merge after @NGTmeaty's review.

Edit: also Copilot made a couple of good suggestions.

NGTmeaty (approved):
LGTM! Thank you!

PR #286, opened 2025-05-19
Optimise XML extractor document loading

vbanos:
The original code loads all the doc in memory before parsing. This is slow and uses 2x memory than needed.

We drop `ReadAll` and pass the body to `xml.NewDecoder()`.


NGTmeaty:
lgtm! Thank you!

NGTmeaty (approved):
lgtm! Thank you!

PR #287, opened 2025-05-19
Increase mimetype detection buffer from 2k to 3k

vbanos:
As suggested by the author of the library for better results. https://github.com/gabriel-vasile/mimetype/blob/66e5c005d80684b64f47eeeb15ad439ee6fad667/mimetype.go#L15

It used to be 2k but they increased it to 3k recently.

### DIFF
--- a/internal/pkg/archiver/body.go
+++ b/internal/pkg/archiver/body.go
@@ -34,9 +34,10 @@ func ProcessBody(u *models.URL, disableAssetsCapture, domainsCrawl bool, maxHops
 		}
 	}
 
-	// Create a buffer to hold the body (first 2KB)
+	// Create a buffer to hold the body (first 3KB) as suggested by mimetype author
+	// https://github.com/gabriel-vasile/mimetype/blob/66e5c005d80684b64f47eeeb15ad439ee6fad667/mimetype.go#L15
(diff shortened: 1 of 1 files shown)


vbanos:
Also note that the library `mimetype` has a `DetectionReader` function that reads the first 3k of a file so that you don't have to make a `buffer` like we do.
https://github.com/gabriel-vasile/mimetype/blob/66e5c005d80684b64f47eeeb15ad439ee6fad667/mimetype.go#L45

I would use that instead of the current code but I hesitated to change it because the current code uses `copyWithTimeoutN` when creating the buffer.
`mimetype` doesn't have a timeout. 
I wonder why we need the timeout, we already have the HTTP response.

CorentinB:
> I would use that instead of the current code but I hesitated to change it because the current code uses `copyWithTimeoutN` when creating the buffer.
> 
> `mimetype` doesn't have a timeout. 
> 
> I wonder why we need the timeout, we already have the HTTP response.
> 
> 

So the timeout in this function is using ReadDeadline which is not configurable at the client level and it's the only way to properly handle "read timeout" when reading HTTP responses.

So yes just changing the buffer size is good! Thanks, lgtm!

CorentinB (approved):

PR #288, opened 2025-05-20
Filter outlinks with unwanted protocols

vbanos:
We filtered some protocols only in the PDF outlink extractor. We didn't catch uppercase and we were missing a lot of protocols.

We implement a common filter function that applies to all extractors and remove the PDF specific code.

We also add a unit test.


CorentinB:
Looks good to me! @NGTmeaty what do you think?

NGTmeaty (approved):
Looks good to me! Thanks! Should reduce the number of entries in HQ!

PR #289, opened 2025-05-20
Define models.NewURL to simplify URL object creation

vbanos:


NGTmeaty (approved):

internal/pkg/source/lq/consumer.go
```diff
@@ -145,15 +145,12 @@ func consumerSender(ctx context.Context, wg *sync.WaitGroup, urlBuffer <-chan *s
...
-				Hops: int(URL.Hops),
-			}
-			err := parsedURL.Parse()
+			parsedURL, err := models.NewURL(URL.Value)
 			if err != nil {
 				discard = true
 			}
-			newItem := models.NewItem(URL.ID, &parsedURL, URL.Via)
+			parsedURL.SetHops(int(URL.Hops))
+			newItem := models.NewItem(URL.ID, parsedURL, URL.Via)
```
yzqzss, line 153: If parsing fails, then it call SetHops() on the nil parsedURL, casue panic.

yzqzss: #294

CorentinB: Good catch!

PR #290, opened 2025-05-21
Regex improvements and add strict regex toggle

NGTmeaty:
This adds the `--strict-regex` toggle which is required to utilize the stricter regex setting.


willmhowes (approved):
lgtm

yzqzss:
I saw this PR touched the part of extracting urls in inline css.

Just a reminder, I will improve this part with standard css parser instead of regex during GSoC.

PR #292, opened 2025-05-21
Make Azure test URLs insensitive to the order of the URL query parameters

yzqzss:
fix failed https://github.com/internetarchive/Zeno/actions/runs/15152201102 test


NGTmeaty (approved):
thanks for the quick fix!

PR #295, opened 2025-05-23
Make terminal logging colorful 🌈

yzqzss:
image

Now my eyes can easily spot the warn/error logs from the terminal stream.

Also added --no-color-logs flag to disable this feature. (to switch back to [slog] handler)

---

This PR only colors the stdout and stderr terminal logging handlers and does not affect the file logging handler (`/jobs/{job}/log/ZENO-{time}.log`).

---

The [slogcolor] handler's output format is slightly different from slog. For example, it does not escape the `"` to `\"` in the log message.

---

Introduced lib:
- https://github.com/MatusOllah/slogcolor


NGTmeaty (approved):
It's so much easier to read. Thank you! LGTM!

CorentinB (approved):
Awesome! LGTM!

PR #304, opened 2025-05-24
Add test with corrupt PDF

vbanos:
Copy InternetArchiveDeveloperPortal.pdf and truncate it to create a corrupt PDF.

Create a PDF unit test with the corrupt PDF and verify that the extractor fails gracefully.


yzqzss (approved):

yzqzss:
LGTM

PR #306, opened 2025-05-26
Increase ProcessBody memory buffer from 2 to 32MB

vbanos:
Avoid disk I/O for temporary WARC files < 32MB to improve performance.


NGTmeaty:
I agree. 32MB also seems too big for a lot of our crawls. A option and/or a larger default would be good with me. 

I'd prefer 6 or 8MB as a good default?

NGTmeaty (approved):
We'll circle back on a CLI option at another time. For now, i think it's fine to merge. Thanks!

PR #307, opened 2025-05-27
NewURL(): return raw URL even if parsing error

yzqzss:
For potential `finisher` debugging needs.

This does not affect any actual behavior. It just makes it more convenient to identify the bad url item that enters `globalHQ.finishCh` when debugging.


NGTmeaty (approved):
Looks good! Thanks!

PR #310, opened 2025-05-27
Handle default HQBatchSize in config

vbanos:
In Go, when a struct field of type int is not explicitly initialized, its default value is 0. So, all config integers default to 0.
Then, in HQ producer we have a special function to set its default value to 100 if its 0.
We can simplify this using viper. We set the default HQBatchSize to 100 in `ConfigInit` and drop the redundant function.

We also add a small unit test to validate config init.

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -138,6 +138,9 @@ func InitConfig() error {
 	once.Do(func() {
 		config = &Config{}
 
+		// Config defaults
+		viper.SetDefault("hq-batch-size", 100)
+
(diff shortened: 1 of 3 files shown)


yzqzss:
https://github.com/internetarchive/Zeno/blob/2dd719dfb238da6a7addc94de79cb24a998b5240/cmd/get_hq.go#L93

The actual default `HQBatchSize` is 500.

Also, setting the default value in `InitConfig()` causes the default value of `getHQCmdFlags()` to be overridden. I don't think it's a good idea to set default values in `InitConfig()`.

PR #314, opened 2025-05-31
Remove unused combinedArgs in FieldedLogger.logWithLevel

vbanos:
The `combinedArgs` is created but its never used anywhere.


yzqzss:
Oh whoops, I forgot to change `record.Add(args...)` to `record.Add(combinedArgs...)`

NGTmeaty:
oops, looks like we both missed it. I assume we want to keep the combined fields - @yzqzss would you like to make a PR?

yzqzss:
> oops, looks like we both missed it. I assume we want to keep the combined fields - @yzqzss would you like to make a PR?

sure! #317

PR #321, opened 2025-06-01
Rewrite NormalizeURL replacing goada with stdlib net/url

vbanos:
The updated `NormalizeURL` passes the same unit tests as the old function.


vbanos:
OK, I understand. It would be good to add some of these URLs in the unit tests of `NormalizeURL` to support this point.

PR #344, opened 2025-06-16
Improve URL mimetype / content-type handling

vbanos:
In some places, we use `URL.GetMIMEType()` and in others we use the HTTP Response `Content-Type` HTTP header.

This isn't consistent and its not right to lookup the `Content-Type` HTTP header in all extractors to define if its JSON, XML, HTML, etc.

We modify `URL` to do that only once and cache it. Then, use `URL.GetMIMEType()` everywhere.

We also add unit tests.

### DIFF
--- a/internal/pkg/archiver/body.go
+++ b/internal/pkg/archiver/body.go
@@ -36,7 +36,7 @@ func ProcessBody(u *models.URL, disableAssetsCapture, domainsCrawl bool, maxHops
 
 	buffer := new(bytes.Buffer)
 	// First check HTTP Content-Type and then fallback to mimetype library.
-	if u.GetMIMEType() != nil {
+	if u.GetMIMEType() == nil {
 		// Create a buffer to hold the body (first 3KB) as suggested by mimetype author
 		// https://github.com/gabriel-vasile/mimetype/blob/66e5c005d80684b64f47eeeb15ad439ee6fad667/mimetype.go#L15
 		if err := copyWithTimeoutN(buffer, u.GetResponse().Body, 3072, conn); err != nil {


yzqzss:
First, the mimetypelib can only perform limited magic detection based on content, and it does not work well with short content and text-based context. The content-based detection is almost impossible to distinguish for many plain text MIME types that do not have magic. For example, `a {background-image: url(“a.jpg”);}` would be identified as `text/plain`.
This PR changes the original comparison between mimeType and contentType to prioritize mimeType over the contentType header, which will at least cause issues with identifying JavaScript and CSS files.

Browsers do use a standard detection process to determine the final MIME type for each resource.

https://mimesniff.spec.whatwg.org/#determining-the-computed-mime-type-of-a-resource

However, the entire process is somewhat complex, and I don't think we need to implement it at this stage. But, generally, the Content-Type header declaration takes precedence over the results of content detection.

Given that our mimetypelib performs poorly in detecting short content and plain text formats in actual archives, we should not use it as the primary source for determining the final MIME type, at least not for plain text content.

Additionally, considering that many HTTP servers do not always return the correct Content-Type header, mimetypelib's detection becomes useful. For example, an HTML file with a Content-Type of text/plain may be detected as text/html by mimetypelib.

---

My opinion is that the current `OR` matching is better, unless we implement behavior similar to a browser's mime sniff and compute functionality.

vbanos:
The main point of the PR is to simplify and make robust the process of determining the mimetype of a `URL` in extractors and other places.
The priority of `Content-Type` vs the `mimetype` library is secondary IMHO.
We use the `Content-Type` to determine the mimetype everywhere except from one place.
https://github.com/internetarchive/Zeno/blob/main/internal/pkg/archiver/body.go#L45

I suggest the following improvement:
`URL.SetMIMEType` is used only there: https://github.com/internetarchive/Zeno/blob/main/internal/pkg/archiver/body.go#L45
We already have the HTTP response headers in the same function.

We modify the code to determine the mimetype using the `Content-Type` header and if its not found we use `mimetype.Detect(buffer.Bytes())` as a fallback.

Everything else remains the same in my PR.
What do you think @yzqzss ?

yzqzss:
Sounds good, let's proceed with that approach :)

pkg/models/url.go
```diff
@@ -75,7 +75,20 @@ func (u *URL) SetDocument(doc *goquery.Document) {
...
 func (u *URL) GetMIMEType() *mimetype.MIME {
+	if u.mimetype != nil {
+		return u.mimetype
+	}
+	if u.GetResponse() != nil {
+		ct := u.GetResponse().Header.Get("Content-Type")
+		if ct != "" {
+			mt := strings.Split(ct, ";")[0]
+			mt = strings.TrimSpace(mt)
+			mt = strings.ToLower(mt)
```
yzqzss: FYI, here we can use the standard library

`mt, params, err := mime.ParseMediaType(ct)`

yzqzss (changes requested):

pkg/models/url.go
vbanos: Thank you for the suggestion, done!

vbanos:
I have updated this PR to address @yzqzss feedback.

internal/pkg/archiver/body.go
```diff
@@ -34,16 +34,17 @@ func ProcessBody(u *models.URL, disableAssetsCapture, domainsCrawl bool, maxHops
 		}
 	}
 
-	// Create a buffer to hold the body (first 3KB) as suggested by mimetype author
-	// https://github.com/gabriel-vasile/mimetype/blob/66e5c005d80684b64f47eeeb15ad439ee6fad667/mimetype.go#L15
 	buffer := new(bytes.Buffer)
-	if err := copyWithTimeoutN(buffer, u.GetResponse().Body, 3072, conn); err != nil {
-		return err
+	// First check HTTP Content-Type and then fallback to mimetype library.
+	if u.GetMIMEType() != nil {
```
yzqzss: `if u.GetMIMEType() == nil {` ?

vbanos: Oh, this was a mistake. Fixed.

PR #351, opened 2025-06-18
Make config more unit-test friendly

vbanos:
The Config struct is stored in a local var and returned via `config.Get()`. If you need to make a unit test that includes e.g. `config.Get().DisableHTMLTag` it is difficult to mock the config object. This is not friendly to unit-testing.

We rename `config` to `AppConfig` in `internal/pkg/config/config.go`. This way, its possible to use `AppConfig.DisableHTMLTag` in other modules and its easy to mock it in unit tests.

To showcase the new syntax, we change the HTML extractor to use the new `AppConfig`.

The old syntax isn't affected.

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -127,7 +127,7 @@ type Config struct {
 }
 
 var (
-	config *Config
+	AppConfig *Config
 	once   sync.Once
 )
 
@@ -137,7 +137,7 @@ var (
 func InitConfig() error {
 	var err error
 	once.Do(func() {
-		config = &Config{}
+		AppConfig = &Config{}
 
 		// Check if a config file is provided via flag
 		if configFile := viper.GetString("config-file"); configFile != "" {
@@ -187,7 +187,7 @@ func InitConfig() error {
 		handleFlagsAliases()
 
 		// Unmarshal the config into the Config struct
-		err = viper.Unmarshal(config)
+		err = viper.Unmarshal(AppConfig)
 	})
 	return err
 }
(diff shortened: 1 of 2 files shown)


vbanos:
If this is merged, I intend to apply the new syntax everywhere we load the config an drop `config.Get()` entirely.

vbanos:
Let's say I want to mock `config.Get().MaxHops`. I would need to add sometlhing like the following in the beginning of a test file (not 100% sure its only that):
```
var appConfig = &Config{
    MaxHops: 5,
}

func Get() *Config {
    return appConfig
}
```
But this would affect all tests in the package. `config.Get()` is a global function. How can I do this for multiple tests? It is confusing.

On the contrary, if I use `AppConfig.MaxHops` with my new suggested syntax, I would have to do the following **inside a test function**. This is much simpler and it doesn't create conflicts with other tests. Also, all the logic is inside the test function and not in a global appConfig object on the top of the test file:
```
    // Save old value to restore later
    originalMaxHops := config.AppConfig.MaxHops
    // Override for test
    config.AppConfig.MaxHops = 1
    defer func() {
        // Restore after test
        config.AppConfig.MaxHops = originalMaxHops
    }()
```

yzqzss:
Whether it is config.Get() or AppConfig, they all point to the same global config pointer. I still don't understand what the renaming does🙈, is it for semantics?

yzqzss:
Oh, didn’t you know that we can actually modify the value of config.Get()? It’s a global pointer after all.

yzqzss:
```go
    // Save old value to restore later
    originalMaxHops := config.Get().MaxHops
    // Override for test
    config.Get().MaxHops = 1
    defer func() {
        // Restore after test
        config.Get().MaxHops = originalMaxHops
    }()
```

Same effect.

vbanos:
OK, in that case, I'm OK with the current approach. Thank you.

PR #352, opened 2025-06-20
fix: missing text/css mimetype

yzqzss:
`gabriel-vasile/mimetype` does not support CSS detection yet, we have to extend a placeholder for our `Content-Type` lookup.


NGTmeaty (approved):
Looks good! Thanks!

PR #353, opened 2025-06-21
Improve GitHub archiving

yzqzss:
Improved the archiving quality of GitHub Issue&PR pages.

26e6dd2ad74420cea746b3b87449fccbf58c8a10, f94197f5947a1030422e86256e398d4f18467af0 fix:
I noticed that some "data:" URLs were being queued that should have been filtered => We forgot to call `filterURLsByProtocol()` for assets.

e543bb3cd6149f060f2a2eb39034e9b0c15a10b0 fix:
GitHub has rewritten the issue pages in react in recent months. Now they embed many useful (images, videos, user avatars...) assets into HTML `<script>` tags as JSON for issue pages.
image

These assets are in JSON strings.
image

and JSON extractor currently only considers an asset if the entire string is a valid URL.

49de5c7e086a34b2ed81ba6581a8fe526069483b improved:
These GitHub assets have no file suffix, so a `sitespecific/github.go` patch was added to JSON extractor to treat them as assets.

daa9ee819535eef58eeb5c5e68de9c722fd2e224 fix:
`hasFileExtension()` wrongly considers _Protocol and domain only, no path_ links like `https://example.com` to have a suffix.

92241c78374d186bad449ecdc6962d7b9143fe93: make `isValidURL()` check more robust

fe3f888 Chnaged: Renamed the original `LinkRegex` to `QuotedLinkRegex` to reflect its actual purpose. The new `LinkRegex` is copyied and adapted from heritrix3. :)

### DIFF
--- a/internal/pkg/postprocessor/extractor/json.go
+++ b/internal/pkg/postprocessor/extractor/json.go
@@ -69,12 +69,22 @@ func findURLs(data interface{}, links *[]string) {
 	case string:
 		if isValidURL(v) {
 			*links = append(*links, v)
+			return
 		} else if isLikelyJSON(v) {
 			// handle JSON in JSON
 			var jsonstringdata interface{}
 			err := json.Unmarshal([]byte(v), &jsonstringdata)
 			if err == nil {
 				findURLs(jsonstringdata, links)
+				return
+			}
+		}
+
+		// find links in text
+		linlsFromText := LinkRegexStrict.FindAllString(v, -1)
+		for _, link := range linlsFromText {
+			if isValidURL(link) {
+				*links = append(*links, link)
 			}
 		}
 	case []interface{}:
--- a/internal/pkg/postprocessor/extractor/json.go
+++ b/internal/pkg/postprocessor/extractor/json.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/ImVexed/fasturl"
+	"github.com/internetarchive/Zeno/internal/pkg/postprocessor/sitespecific/github"
 	"github.com/internetarchive/Zeno/pkg/models"
 )
 
@@ -44,7 +45,7 @@ func GetURLsFromJSON(decoder *json.Decoder) (assets, outlinks []string, err erro
 
 	// We only consider as assets the URLs in which we can find a file extension
 	for _, link := range links {
-		if hasFileExtension(link) {
+		if hasFileExtension(link) || github.ShouldConsiderAsAsset(link) {
 			assets = append(assets, link)
 		} else {
 			outlinks = append(outlinks, link)
--- a/internal/pkg/postprocessor/extractor/json.go
+++ b/internal/pkg/postprocessor/extractor/json.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/ImVexed/fasturl"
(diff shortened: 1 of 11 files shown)


yzqzss:
Before (left window), after (right window):

https://github.com/user-attachments/assets/5750a78e-ca9a-48a4-89d0-87c6d553e80a

You can see that the avatars, pictures, and video attachment have been captured correctly.

Example issue url: `https://github[.]com/webrecorder/browsertrix-behaviors/issues/58`

yzqzss:
ready

internal/pkg/postprocessor/extractor/json.go
```diff
@@ -69,12 +70,22 @@ func findURLs(data interface{}, links *[]string) {
...
 			var jsonstringdata interface{}
 			err := json.Unmarshal([]byte(v), &jsonstringdata)
 			if err == nil {
 				findURLs(jsonstringdata, links)
+				return
+			}
+		}
+
+		// find links in text
+		linksFromText := LinkRegexStrict.FindAllString(v, -1)
```
NGTmeaty: LinkRegexStrict was very heavy last I looked at our Pyroscope, we even [added a setting](https://github.com/internetarchive/Zeno/blob/f7a845339a2f8524b9a453a79bab51af8e12073d/internal/pkg/postprocessor/outlinks.go#L120) to only use it when needed. 

Do you think we need to use LinkRegexStrict or would LinkRegex also work? Do you think we could use the setting here?

internal/pkg/postprocessor/sitespecific/github/github.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/internetarchive/Zeno/internal/pkg/log"
+)
+
+var logger = log.NewFieldedLogger(&log.Fields{
+	"component": "postprocessor.sitespecific.github",
+})
+
+// user avatars
+var githubAvatar = regexp.MustCompile(`(?i)^https://avatars\.githubusercontent\.com/u/`)
```
NGTmeaty: I wonder if it would make sense to replace `githubAvatar`, `githubFrontendAssets`, and `githubPrivateUserImages` with a check instead for the domains `githubusercontent.com` and `githubassets.com`? In case the specific paths change in the future or other assets that could be included on these domains?

NGTmeaty (commented):
Thank you! These improvements look super solid and very helpful! I had a couple of small questions in my review:

internal/pkg/postprocessor/sitespecific/github/github.go
yzqzss: done 9b30336ef378d1ee28f2215ad4876280da9892e8

internal/pkg/postprocessor/extractor/json.go
yzqzss: Thank you for letting me know!

done b30b34e2177366b52999ca6f28af1d8a29b1d294

yzqzss:
Also renamed the original `LinkRegex` to QuotedLinkRegex to reflect its actual purpose.
The new `LinkRegex` is copyied and adapted from heritrix3. :)

fe3f88869af9f4966bec4f443404ff615bdd8da6

NGTmeaty (approved):
Thanks! Changes look great!

PR #356, opened 2025-06-24
Headless/Headful mode

yzqzss:
close: #347


CorentinB:
Incredible work!

yzqzss:
I'll add PR details later.

go.mod
```diff
@@ -12,6 +12,8 @@ require (
 	github.com/fatih/color v1.18.0
 	github.com/gabriel-vasile/mimetype v1.4.9
 	github.com/gdamore/tcell/v2 v2.8.1
+	github.com/go-rod/rod v0.116.2
+	github.com/go-rod/stealth v0.4.9
```
NGTmeaty, line 17: go-rod/stealth hasn't been updated in nearly 2 years, does this still provide functionality that we find useful?

internal/pkg/archiver/headless/archiver.go
```diff
@@ -0,0 +1,241 @@
...
+	warc "github.com/internetarchive/gowarc"
+)
+
+//go:embed behaviors.js
+var behaviorsJS string
+
+func behaviorInitJS() string {
+	options := map[string]any{
+		"autoscroll":   false,
+		"autofetch":    false, // disabled by default, this function will fetch resources twice.
```
NGTmeaty, line 40: Could we log a warning when this flag is enabled to let the operator know resources will be fetched twice?

internal/pkg/archiver/headless/archiver.go
```diff
@@ -0,0 +1,241 @@
...
+
+func clientDo(h *rod.Hijack, client *http.Client) (*http.Response, error) {
+	resp, err := client.Do(h.Request.Req())
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: handle hijack response in another function
+	h.Response.Payload().ResponseCode = resp.StatusCode
+	h.Response.RawResponse = resp
```
NGTmeaty, line 79: I was initially a little confused - but I do see now that the client being used is the warc HTTP client and the response handling here is to make it available to go-rod. I know that ArchiveTeam has had issues with CDP before (examples [here](https://wiki.archiveteam.org/index.php/Chromebot#Shutdown)) and I would want to be sure to avoid those, so I will have to spend some additional time on validation here, but I do like the look of things so far!

internal/pkg/archiver/headless/archiver.go
```diff
@@ -0,0 +1,241 @@
...
+		"seed_id":   seed.GetShortID(),
+		"item_id":   item.GetShortID(),
+		"url":       item.GetURL().String(),
+	})
+
+	// Set the hijack router
+	router := HeadlessBrowser.HijackRequests()
+	defer router.MustStop()
+
+	flyingRequests := NewWaitGroup()
```
NGTmeaty: suggested change: inflightRequests?

internal/pkg/archiver/headless/archiver.go
```diff
@@ -0,0 +1,241 @@
...
+		logger := log.NewFieldedLogger(&log.Fields{
+			"component": "archiver.archiveHeadless.router",
+			"seed_id":   seed.GetShortID(),
+			"item_id":   item.GetShortID(),
+			"url":       hijack.Request.URL().String(),
+		})
+
+		// drop requests that are not in the allowed methods
+		if !slices.Contains(config.Get().HeadlessAllowedMethods, hijack.Request.Method()) {
+			logger.Debug("droppinp request not in allowed methods", "method", hijack.Request.Method())
```
NGTmeaty: nit: "dropping"

cmd/get.go
```diff
@@ -54,6 +54,24 @@ func getCMDsFlags(getCmd *cobra.Command) {
 	getCmd.PersistentFlags().Float64("min-space-required", 0, "Minimum space required in GB to continue the crawl. Default will be 50GB * (total disk space / 256GB) if total disk space is less than 256GB, else 50GB.")
 	getCmd.PersistentFlags().Bool("strict-regex", false, "If turned on, the xurls `strict` regex setting will be used. Otherwise a looser regex will be used.")
 
+	// Headless flags
+	getCmd.PersistentFlags().Bool("headless", false, "[headless] Run in headless mode (experimental)")
+	getCmd.PersistentFlags().Bool("headfull", false, "[headless] Run in headfull mode")
```
NGTmeaty: Perhaps a small explanation for what 'headfull' actually does in case it isn't obvious?  (and that both flags are required to start it?)

NGTmeaty (commented):
Looks great so far! I have a couple of small comments but primarily I need to do some validation of the final WARC to ensure all relevant headers are being recorded correctly!

cmd/get.go
```diff
@@ -54,6 +54,24 @@ func getCMDsFlags(getCmd *cobra.Command) {
...
+	getCmd.PersistentFlags().Bool("headless-trace", false, "[headless] Enable headless trace mode")
+	getCmd.PersistentFlags().Bool("headless-dev-tools", false, "[headless] F12")
+	getCmd.PersistentFlags().Bool("headless-stealth-mode", false, "[headless] Run in stealth mode and use the browser's native User-Agent")
+	getCmd.PersistentFlags().Bool("headless-user-mode", false, "[headless] Launch your regular user browser session")
+
+	getCmd.PersistentFlags().String("headless-chromium-bin", "", "[headless] Bin of the browser binary path to launch, if the path is not empty the auto download will be disabled.")
```
NGTmeaty: Flagging something: go-rod by default downloads Chrome 128. This is _very_ old and I would be concerned to use this in production.

(Revision number is calculated here https://github.com/go-rod/rod/blob/main/lib/launcher/revision.go#L6 )

go.mod
yzqzss: Yes, it is still useful. It can indeed bypass some simple 0click bot checks.
For example, to archive https://zhuanlan.zhihu.com/p/20062978, I have to use a newer browser version + browser UA + headfull + stealth.

internal/pkg/archiver/headless/archiver.go
yzqzss: Done

internal/pkg/archiver/headless/archiver.go
yzqzss: done

yzqzss:
outlinks crawling (--max-hops > 0) now works, but not reliably. :(

https://github.com/user-attachments/assets/abd2fe35-afb0-44da-9387-b3d05110a8a5

internal/pkg/archiver/headless/archiver.go
yzqzss: 60c27c7ff218e809f56c648475a7e1102855af6f

The headless WARCs are marked so that we can exclude them if we find problem in the future.

cmd/get.go
yzqzss: 71c47bbd25c651119d189640698cd08ec1bd76be

yzqzss:
> Can you please describe exactly what's unreliable?

Ahhhh, the rate limit on githubassets.com misled me. (Because I included a lot of github urls when testing)
If there are too many requests in a short period of time, it will not respond to any TCP connections. Makes me feel that headless is sometimes unstable.

---

Since we hijacked the browser's request, the browser's built-in cache was not used. Each page has to re-download cacheable page resources (js, css, images...), resulting in many repeated requests...
I don't have a solution for this at the moment...

NGTmeaty:
> Since we hijacked the browser's request, the browser's built-in cache was not used. Each page has to re-download cacheable page resources (js, css, images...), resulting in many repeated requests...
> I don't have a solution for this at the moment...

I can't remember if I created an issue for this - but we wanted to add some sort of URL counter to Zeno overall as well to ensure that we aren't downloading assets hundreds of times - even if we are writing revisit records. But that is sort of outside of the scope of this PR. I'll go double check if we did have one.

internal/pkg/archiver/headless/body.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	fullBody, err := processBodyHeadless(u)
+	return fullBody, body.CloseConnWithError(bodyLogger, conn, err)
+}
+
+func processBodyHeadless(u *http.Response) ([]byte, error) {
+	onExit := atomic.Bool{}
+
+	if os.Getenv("ZENO_DEBUG_HEADLESS_TRACE") != "" {
```
NGTmeaty: Is this something we want to leave in? If so - can we document it?

internal/pkg/archiver/headless/browser.go
```diff
@@ -0,0 +1,99 @@
...
+var Launcher *launcher.Launcher
+
+var DefaultChromiumRevision = 1465706
+
+var browserLogger = log.NewFieldedLogger(&log.Fields{"component": "archiver.headless.client"})
+
+func queryLatestChromiumRevision() (int, error) {
+	client := &http.Client{Timeout: 10 * time.Second}
+	// From web page: https://chromiumdash.appspot.com/releases?platform=Linux
+	resp, err := client.Get("https://chromiumdash.appspot.com/fetch_releases?channel=Stable&platform=Linux&num=1&offset=0")
```
NGTmeaty: I know we want to be as up to date as possible - but is there any concern about losing compatibility with go-rod? Do we currently know what would happen if the CDP connection fails or something similar?

internal/pkg/archiver/headless/bx_logger.go
```diff
@@ -0,0 +1,65 @@
...
+	suppressedMessages int
+}
+
+func newBxLogger(item *models.Item) *ZenoBxLogger {
+	return &ZenoBxLogger{
+		item: item,
+	}
+}
+
+// LogFunc expose itself to browser, so we
```
NGTmeaty: Comment looks like it was cut off

internal/pkg/archiver/warc.go
```diff
@@ -17,9 +18,14 @@ func startWARCWriter() {
 	rotatorSettings.WarcSize = float64(config.Get().WARCSize)
 	rotatorSettings.OutputDirectory = path.Join(config.Get().JobPath, "warcs")
 
+	version := utils.GetVersion()
+	rotatorSettings.WarcinfoContent.Set("software", "Zeno/"+version.Version+" warc/"+version.WarcVersion)
```
NGTmeaty, line 22: I could have sworn we were setting this somewhere but I guess not. Thank you!

yzqzss: We have ~~always~~ (if `--user-agent` or `--headless-stealth` flag not on) included Zeno version information in `User-Agent` request header.

internal/pkg/archiver/headless/browser.go
yzqzss: > but is there any concern about losing compatibility with go-rod

https://chromedevtools.github.io/devtools-protocol/1-3/

Currently, Chromium uses DevTools-Protocol v1.3. As long as the protocol version remains unchanged, all Chromium versions _should_ be compatible.

Added a panic as a canary: be1b5aa25c4e29078470a86ad1e8be95d7648f53

internal/pkg/archiver/headless/body.go
yzqzss: cleaned

internal/pkg/archiver/headless/browser.go
NGTmeaty: Perfect. Thanks!

yzqzss:
> > Can you please describe exactly what's unreliable?
> 
> Ahhhh, the rate limit on githubassets.com misled me. (Because I included a lot of github urls when testing) If there are too many requests in a short period of time, it will not respond to any TCP connections. Makes me feel that headless is sometimes unstable.

Ahhhhhhh, this's not the root cause.
The root cause was that the `conntrack` configuration of my network outbound device was too small...

```
dmesg: [1219292.731593] nf_conntrack: nf_conntrack: table full, dropping packet
```

```
~# cat /proc/sys/net/netfilter/nf_conntrack_max
4096
```

gowarc new TUP connection for each request, while conntrack keeps the connection information in it's internal table for a while after it is closed. When the table is full, Linux rejects any new connections.

CorentinB:
> gowarc new TCP connection for each request
Reminds me that @equals215 and I explored a path for improvement related to that a while ago, without much success.. Might be worth having fresh eyes to look at this problem

cmd/get.go
```diff
@@ -54,21 +68,44 @@ func getCMDsFlags(getCmd *cobra.Command) {
...
+
+func addHeadlessFlags(getCmd *cobra.Command) {
+	getCmd.PersistentFlags().Bool("headless", false, "[headless] Run in headless mode (experimental).")
+	getCmd.PersistentFlags().Bool("headless-headful", false, "[headless] Switch to headful mode to run the browser with full graphics stack and show browser GUI (requires --headless).")
+	getCmd.PersistentFlags().Bool("headless-trace", false, "[headless] Trace enables the visual tracing of the input actions on the page.")
+	getCmd.PersistentFlags().Bool("headless-dev-tools", false, "[headless] F12")
+	getCmd.PersistentFlags().Bool("headless-stealth", false, "[headless] Use stealth to prevent bot detection and use the browser's native User-Agent")
+	getCmd.PersistentFlags().Bool("headless-user-mode", false, "[headless] Launch your regular user browser session")
+
+	getCmd.PersistentFlags().Int("headless-chromium-revision", headless.DefaultChromiumRevision, "[headless] Default Chromium binary revision to auto download and use (if --headless-chromium-bin is not set). set to -1 to query the latest revision from chromiumdash.appspot.com")
```
NGTmeaty: I think I would want to see this the other way round - set to -1 by default to ensure we are getting the latest version (with the "default" binary in the description)

yzqzss: Done.

Note that the Chromium stable channel is released every week.
we probably need to clean up `~/.cache/rod/browser/` from time to time.

yzqzss:
Ready for review (1) (1) (1)
😅

yzqzss:
image

https://storage.googleapis.com/chromium-browser-snapshots/Linux_x64/1477651/chrome-linux.zip (404)
E2E failed, Chromium hasn't build snapshot for revision 1477651, not sure why.

internal/pkg/postprocessor/extractor/link_header_test.go
```diff
@@ -113,19 +113,19 @@ func TestExtractURLsFromHeader(t *testing.T) {
...
 			}
 
-			err = archiver.ProcessBody(URL, false, false, 0, os.TempDir())
+			err = generalarchiver.ProcessBody(URL, false, false, 0, os.TempDir(), nil)
 			if err != nil {
 				t.Errorf("ProcessBody() error = %v", err)
 			}
 
-			got := ExtractURLsFromHeader(URL)
+			got, _ := LinkHeaderExtractor{}.Extract(URL)
```
NGTmeaty: Check for err as well?

internal/pkg/postprocessor/outlinks.go
```diff
@@ -31,19 +32,34 @@ var outlinkExtractors = []OutlinkExtractor{
...
+		contentType string
 		logger      = log.NewFieldedLogger(&log.Fields{
 			"component": "postprocessor.extractOutlinks",
 		})
 	)
 
+	if item.GetURL().GetResponse() != nil {
+		contentType = item.GetURL().GetResponse().Header.Get("Content-Type")
+	} else {
+		contentType = "text/html" // Headless, hardcoded to HTML
```
NGTmeaty, line 44: Are we only receiving HTML here? If not - should we be expecting to parse other types as well?

internal/pkg/archiver/headless/archiver.go
```diff
@@ -0,0 +1,453 @@
...
+	await self.__bx_behaviors.run();
+}`
+
+func clientDo(client *http.Client, req *http.Request, h *rod.Hijack) (*http.Response, error) {
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: handle hijack response in another function
```
NGTmeaty: Is this TODO still relevant?

internal/pkg/archiver/headless/archiver.go
```diff
@@ -0,0 +1,453 @@
...
+
+		logger := log.NewFieldedLogger(&log.Fields{
+			"component": "archiver.headless.router",
+			"item_id":   item.GetShortID(),
+			"seed_id":   seed.GetShortID(),
+			"item_url":  item.GetURL().String(),
+			"url":       hijack.Request.URL().String(),
+		})
+
+		// debug:
```
NGTmeaty: Make sure this gets removed please! :)

internal/pkg/postprocessor/extractor/link_header_test.go
yzqzss: This function does not actually return error.

The redundant error return value has been deleted. 2502b2942e360b953201a4bc19bf40de1d699d2b

internal/pkg/postprocessor/outlinks.go
yzqzss: Everything is html page (more precisely, DOM tree) in browser, Chromium render other previewable non-HTML resources in HTML.

for example, if I open a .mp4 url in tab, browser will create a HTML container to render the mp4:

```
<html>
<head><meta name="viewport" content="width=device-width"></head>
<body><video controls="" autoplay="" name="media"><source src="{URL}" type="video/mp4"></video></body>
</html>
```

if open a .txt url:

```
<html><head><meta name="color-scheme" content="light dark"></head><body><pre style="word-wrap: break-wo
rd; white-space: pre-wrap;">
...
</pre></body></html>
```

----

image

Also, if the page tab cannot be represented as HTML, the item will be marked as `ItemFailed` and the post-processing (the outlink extractor) will be skipped.

NGTmeaty: Hmm, maybe I am a little confused - are we getting both the raw bytes and the Chromium processed bytes? From my current tests it seems like we are recording the raw bytes to WARC but is this something we have to be careful of?

yzqzss: yes, raw bytes go to WARC.

I have to use Chromium processed DOM html as `item.body` for post-processing. This is because the web page DOM is not necessarily composed from one specific request, especially when the web page is dynamic and driven by JS.

yzqzss:
So, let me pin the default chromium revision first.

I think in production environment, we can use the release version of chromium binary from OS distribution. (`--headless-chromium-bin`)

internal/pkg/postprocessor/item.go
```diff
@@ -28,7 +28,7 @@ func postprocessItem(item *models.Item) []*models.Item {
 	logger.Debug("postprocessing item")
 
 	// Verify if there is any redirection
-	if isStatusCodeRedirect(item.GetURL().GetResponse().StatusCode) {
+	if item.GetURL().GetResponse() != nil && isStatusCodeRedirect(item.GetURL().GetResponse().StatusCode) {
```
willmhowes, line 31: I see many instances of this `nil` check which I understand to be implicitly checking whether or not the crawl mode is headless or standard. I think we would benefit from having a README (either in the main README or headless README) that describes what postprocessing is handled by other processes in headless mode vs needs to be manually extracted when crawling in standard mode

internal/pkg/archiver/body/body.go
willmhowes, line 1: Why do we have `internal/pkg/archiver/body/body.go` when we could merge it with the file `internal/pkg/archiver/general/body.go` ? In any case, I feel like there is one too many files in the `internal/pkg/archiver` directory named `body.go` so maybe we could at least rename `internal/pkg/archiver/body/body.go` to be more specific about the connection logic it contains.

willmhowes, line 1: Why does this `body.go` file have it's own `body` directory?

willmhowes (changes requested):
This is really solid work @yzqzss! I'm still wondering if there is room to improve the structure of the `internal/pkg/archiver` directory so that it's more clear what code drives the "standard mode" crawls vs "headless mode" crawls, but overall very impressed

internal/pkg/postprocessor/item.go
yzqzss: Added more notes to the headless README.

Please feel free to tell me anything that can be improved.

internal/pkg/archiver/body/body.go
yzqzss: This is mainly to avoid circular imports and to decouple headless code from general code.

This separate `body` package mainly handles connection-related matters. Renamed it to `connutil` (the better name `conn` is already used by variables). ab5b482ebe1a03e148c705163018c36398639cad

NGTmeaty (approved):
Looks great! Thanks again for your hard work here!

PR #365, opened 2025-06-26
Remove redundant code in domainscrawl

vbanos:
This code checks if `globalMatcher.enabled` is false and it sets it to true. This means that `globalMatcher.enabled` should be always just true.


NGTmeaty (approved):
Looks good. Thank you!

PR #370, opened 2025-06-27
handle non-utf8 html document and their URLs

### LINKED ISSUE #169: Handle non-UTF8 HTML pages
https://github.com/PuerkitoBio/goquery/wiki/Tips-and-tricks

yzqzss:
close #169 

Some temporary references:

---

HTML:

html encoding:
https://encoding.spec.whatwg.org (spec, en)
→ https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding (spec, en)
→ https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-charset (spec, en)
https://en.wikipedia.org/wiki/Windows-1252 (wiki, en)
> websites declared ISO 8859-1 which is treated as Windows-1252 by all modern browsers (as required by the HTML5 standard) → https://w3techs.com/technologies/overview/character_encoding

https://www.xiaogd.net/md/web-page-encoding-and-mojibake-1 (blog, zh)
https://httptoolkit.com/blog/http-encodings/ (blog, en)

---

URL:

https://html.spec.whatwg.org/multipage/urls-and-fetching.html#resolving-urls

> Let encoding be UTF-8.
> If environment is a Document object, then set [encoding] (document charset) to environment's character encoding.

https://url.spec.whatwg.org/#query-state

> 2. Percent-encode after encoding, with [encoding], buffer, and queryPercentEncodeSet, and append the result to url’s query.

https://url.spec.whatwg.org/#path-state

> 1. ...special cases...
> 2. Otherwise, run these steps:
>    .... special cases ...
>    3. UTF-8 percent-encode c using the path percent-encode set and append the result to buffer.


internal/pkg/postprocessor/extractor/html_document.go
```diff
@@ -0,0 +1,148 @@
...
+	return r, nil, enc, name, certain
+}
+
+// TransformDocument transforms the document of a URL by detecting its encoding and creating a utf-8 goquery document.
+func TransformDocument(u *models.URL) (doc *goquery.Document, err error) {
+	u.DocumentTransformMux.Lock()
+	defer u.DocumentTransformMux.Unlock()
+
+	// debug: reset cache
+	// u.SetDocumentCache(nil)
```
NGTmeaty: Remove comment please :)

internal/pkg/postprocessor/extractor/html_document.go
```diff
@@ -0,0 +1,148 @@
...
+
+		r = transform.NewReader(r, enc.NewDecoder())
+	}
+
+	return r, nil, enc, name, certain
+}
+
+// TransformDocument transforms the document of a URL by detecting its encoding and creating a utf-8 goquery document.
+func TransformDocument(u *models.URL) (doc *goquery.Document, err error) {
+	u.DocumentTransformMux.Lock()
```
NGTmeaty: Is anything else expected to be using the document while it's being transformed? I can't see another use of the mutex?

e2e/test/nonutf8encoding/nonutf8encoding_test.go
```diff
@@ -0,0 +1,112 @@
...
+			w.Write(gbkPayload)
+			return
+		}
+		http.NotFound(w, r)
+	}))
+}
+
+func TestNonUTF8Encoding(t *testing.T) {
+	server := setupServer()
+	serverURL := strings.Replace(server.URL, "127.0.0.1", "127.0.0.1.nip.io", 1)
```
NGTmeaty: Is there a reason why we need to use `nip.io` here?

e2e/test/nonutf8encoding/nonutf8encoding_test.go
```diff
@@ -0,0 +1,112 @@
...
+			if strings.Contains(r.URL.RawQuery, "%CA%C0%BD%E7=%D4%D9%BC%FB") { // >>> '世界=再见'.encode('gbk') = b'\xca\xc0\xbd\xe7=\xd4\xd9\xbc\xfb'
+				w.Header().Set("Content-Type", "text/plain")
+				w.Write([]byte("OK"))
+				return
+			} else {
+				http.Error(w, "Bad Request - bad query", http.StatusBadRequest)
+				return
+			}
+		case "/":
+			w.Header().Set("Content-Type", "text/html; charset=gbk") // Declare GBK encoding
```
NGTmeaty: We also support determining the encoding - could we also test with that?

NGTmeaty (commented):
Couple of small comments otherwise looks great!

internal/pkg/postprocessor/extractor/html_document.go
yzqzss: I had previously assumed that the outlink extrator and asset extractor would run simultaneously. just checked, they are executed sequentially and indeed no locks are needed here.

e2e/test/nonutf8encoding/nonutf8encoding_test.go
yzqzss: To bypass the localhost check... a bit hacky, yes.

https://github.com/internetarchive/Zeno/blob/5f6a4be74bc5d67ee747c761cf7ece87cdc8b96f/internal/pkg/preprocessor/url.go#L60-L62

NGTmeaty: Ahhhhh. I suppose it's not a big deal but maybe a flag to allow this. No worries though. Thank you!

NGTmeaty (approved):
Looks great! Thank you for the changes and tests!

Issue #245, opened 2025-03-06
Multi-platform builds (Linux, Darwin, Windows) and architectures (amd64, arm64)


karanngi:
Add a Goreleaser configuration for multi-platform builds, enabling users to seamlessly run the `Zeno` on any platform (Linux, Darwin, Windows) and architecture (amd64, arm64) of their choice.

https://goreleaser.com/

CorentinB:
Seems like a fine idea, although we should be more clear that we intend Zeno to be used on Linux, and not really on anything else.

Ilurusheshasai:
How can I contribute to this issue and work hand in hand with you?

shirsakm:
@CorentinB I was thinking of submitting my GSoC application for the Zeno idea. The idea given is kinda vague, but right now I wanna familiarize myself with the codebase, do you think this would be a good first issue for that, or should I tackle something like #159, #160 or #161, are those issues more pertinent?

yzqzss:
There are currently only two factors that prevent Zeno from building and running on Windows and other OS:
1. Zeno's diskwatcher only uses linux syscall to get the free space size of the PWD.
2. warclib's DNS is hardcoded to read from /etc/resolv.conf. On some Linux distributions/containers/OSs such as Windows, nixos, termux, etc., this path does not exist.

yzqzss:
Providing binaries for other platforms does help people try Zeno :)
Although, like CorentinB said, Zeno targets standard Linux, and we cannot guarantee that it will work properly in other environments.

CorentinB:
> @CorentinB why reopen this issue? Multi-platform builds are working if you look here - https://github.com/internetarchive/Zeno/releases/tag/v2.0.15

No actually it was opened, I closed it then realised I wasn't sure so reopened it lol

PR #374, opened 2025-07-07
refactor(preprocessor): remove url normalizer with cpp bindings

otkd:
Step towards multiplatform, in this case simpler build on mac

#245


internal/pkg/preprocessor/url.go
```diff
@@ -15,48 +22,57 @@ func NormalizeURL(URL *models.URL, parentURL *models.URL) (err error) {
...
+			}
+		}
 	}
 
-	adaParse.SetHash("")
-	if scheme := adaParse.Protocol(); scheme != "http:" && scheme != "https:" {
+	wuParse.SetHash("")
+
+	scheme := strings.ToLower(wuParse.Protocol())
+	if scheme != "http:" && scheme != "https:" {
```
yzqzss: https://github.com/nlnwa/whatwg-url/blob/605b909fe66d0472311892f4feed6b4de3e2bdb8/url/parser.go#L171

whatwg-url has normalized the scheme to lowercase.

```suggestion
	scheme := wuParse.Scheme()
	if scheme != "http" && scheme != "https" {
```
```

File: internal/pkg/preprocessor/url.go [hunk 2]
```diff
@@ -66,8 +82,8 @@ func NormalizeURL(URL *models.URL, parentURL *models.URL) (err error) {
 		return ErrUnsupportedHost
 	}
 
-	URL.Raw = adaParse.Href()
-	adaParse.Free()
+	// Update the URL with the normalized version
+	URL.Raw = wuParse.Href(false)
```
yzqzss:
```suggestion
	URL.Raw = wuParse.Href(true /* exclude fragment */)
```

and remove the `wuParse.SetHash("")` above.

internal/pkg/preprocessor/url.go
```diff
@@ -15,48 +22,57 @@ func NormalizeURL(URL *models.URL, parentURL *models.URL) (err error) {
...
+				return err
+			}
+		} else {
+			baseURL := parentURL.GetParsed()
+			if baseURL == nil {
+				return fmt.Errorf("invalid baseURL in parentURL: %s", parentURL.Raw)
+			}
+
+			resolved := baseURL.ResolveReference(parsedURL)
+			wuParse, err = wu.Parse(resolved.String())
```
yzqzss:
```suggestion
			// Assert that parentURL is parsed
			if parentURL.GetParsed() == nil {
				return fmt.Errorf("parent URL is not parsed: %s", parentURL.Raw)
			}

			wuParse, err = wu.ParseRef(parentURL.String() /* base */, URL.Raw /* ref */)
```

internal/pkg/preprocessor/url.go
```diff
@@ -15,48 +22,57 @@ func NormalizeURL(URL *models.URL, parentURL *models.URL) (err error) {
...
+			if err != nil {
+				return err
+			}
+		}
 	}
 
-	adaParse.SetHash("")
-	if scheme := adaParse.Protocol(); scheme != "http:" && scheme != "https:" {
+	wuParse.SetHash("")
```
yzqzss:
```suggestion
```

yzqzss (changes requested):

yzqzss:
Could you please append these test cases to url_test.go?
:)

```
		{
			name:        "URL with fragment",
			rawURL:      "https://example.com/path#fragment",
			wantErr:     false,
			expectedURL: "https://example.com/path",
		},
		{
			name:        "Unicode URL",
			rawURL:      "https://通知.收到了/Zeno/崛起.html?query=了",
			wantErr:     false,
			expectedURL: `https://xn--k7y792c.xn--ykq12govw/Zeno/%E5%B4%9B%E8%B5%B7.html?query=%E4%BA%86`,
		},
		{
			name:        "Uppercase Scheme and Domain URL",
			rawURL:      "HTTPS://EXAMPLE.COM/PATH",
			wantErr:     false,
			expectedURL: "https://example.com/PATH",
		},
		{
			name:        "URL with dots directory in path",
			rawURL:      "https://example.com/../../1/2/../3/./4/5",
			wantErr:     false,
			expectedURL: "https://example.com/1/3/4/5",
		},
		{
			name:        "Localhost URL",
			rawURL:      "http://localhost/path",
			wantErr:     true,
			expectedURL: "",
		},
		{
			name:        "URL with semicolon separator in query",
			rawURL:      "https://example.com/path?q1=a;q2=b",
			wantErr:     false,
			expectedURL: "https://example.com/path?q1=a;q2=b",
		},
```

willmhowes (changes requested):
The reason why Zeno uses `goada` is explained [here](https://github.com/internetarchive/Zeno/pull/321#issuecomment-2932059712). BUT, I'm happy to move forward without `goada` when
- [ ] your pull request can pass the tests proposed by @yzqzss
- [ ] you can confirm that `nlnwa/whatwg-url` has at least similar performance to `goada`. I'm expecting this to be the case, but want to be sure

internal/pkg/preprocessor/url.go
```diff
@@ -15,48 +22,57 @@ func NormalizeURL(URL *models.URL, parentURL *models.URL) (err error) {
...
+			lowerURL := strings.ToLower(URL.Raw)
+			if !strings.HasPrefix(lowerURL, httpPrefix) &&
+				!strings.HasPrefix(lowerURL, httpsPrefix) &&
+				!strings.HasPrefix(lowerURL, ftpPrefix) &&
+				!strings.Contains(lowerURL, "://") {
+				URL.Raw = httpPrefix + URL.Raw
 			}
-		} else {
-			adaParse, err = goada.NewWithBase(URL.Raw, baseURL.String())
+			wuParse, err = wu.Parse(URL.Raw)
```
yzqzss:
```
			if ... {
				URL.Raw = httpPrefix + URL.Raw
				wuParse, err = wu.Parse(URL.Raw)
			}
			if err != nil {
				return err
			}
```

Move `wuParse, err = wu.Parse(URL.Raw)` into the `if` block so we don’t parse twice on `false` condition.

yzqzss:
```go
func BenchmarkNormalizeURL(b *testing.B) {
        tests ... // copy from TestNormalizeURL()
	for _, tc := range tests {
		b.Run(tc.name, func(b *testing.B) {
			var parentURL *models.URL
			if tc.parentURL != "" {
				parentURL = &models.URL{Raw: tc.parentURL}
				parentURL.Parse()
			}

			b.ResetTimer()
			for i := 0; i < b.N; i++ {
				url := &models.URL{Raw: tc.rawURL}
				err := NormalizeURL(url, parentURL)
				if tc.wantErr && err == nil {
					b.Errorf("NormalizeURL() expected error but got none")
				} else if !tc.wantErr && err != nil {
					b.Errorf("NormalizeURL() error = %v", err)
				}
			}
		})
		b.Run(tc.name+"(OLD)", func(b *testing.B) {
			var parentURL *models.URL
			if tc.parentURL != "" {
				parentURL = &models.URL{Raw: tc.parentURL}
				parentURL.Parse()
			}

			b.ResetTimer()
			for i := 0; i < b.N; i++ {
				url := &models.URL{Raw: tc.rawURL}
				err := NormalizeURLOld(url, parentURL)
				if tc.wantErr && err == nil {
					b.Errorf("NormalizeURL() expected error but got none")
				} else if !tc.wantErr && err != nil {
					b.Errorf("NormalizeURL() error = %v", err)
				}
			}
		})
	}
}

```

```bash
go test -run='^$' -benchmem -benchtime=100000x -bench 'BenchmarkNormalizeURL' github.com/internetarchive/Zeno/internal/pkg/preprocessor
```


```
BenchmarkNormalizeURL/valid_absolute_URL-12                                 100000              2303 ns/op             960 B/op         40 allocs/op
BenchmarkNormalizeURL/valid_absolute_URL(OLD)-12                            100000              1347 ns/op             440 B/op          8 allocs/op
BenchmarkNormalizeURL/valid_relative_URL_with_parent-12                     100000              2847 ns/op            1616 B/op         52 allocs/op
BenchmarkNormalizeURL/valid_relative_URL_with_parent(OLD)-12                100000              1280 ns/op             368 B/op          7 allocs/op
BenchmarkNormalizeURL/invalid_URL-12                                        100000               362.7 ns/op           312 B/op          6 allocs/op
BenchmarkNormalizeURL/invalid_URL(OLD)-12                                   100000               120.2 ns/op           208 B/op          3 allocs/op
BenchmarkNormalizeURL/valid_URL_without_scheme-12                           100000              3203 ns/op            1920 B/op         69 allocs/op
BenchmarkNormalizeURL/valid_URL_without_scheme(OLD)-12                      100000              1125 ns/op             421 B/op          8 allocs/op
BenchmarkNormalizeURL/FTP_url-12                                            100000              1679 ns/op             632 B/op         25 allocs/op
BenchmarkNormalizeURL/FTP_url(OLD)-12                                       100000              1965 ns/op             236 B/op          5 allocs/op
BenchmarkNormalizeURL/valid_URL_with_path_without_scheme-12                 100000              3537 ns/op            2016 B/op         76 allocs/op
BenchmarkNormalizeURL/valid_URL_with_path_without_scheme(OLD)-12            100000              1202 ns/op             445 B/op          8 allocs/op
BenchmarkNormalizeURL/URL_with_leading_and_trailing_quotes-12               100000              2272 ns/op             960 B/op         40 allocs/op
BenchmarkNormalizeURL/URL_with_leading_and_trailing_quotes(OLD)-12          100000              1269 ns/op             440 B/op          8 allocs/op
BenchmarkNormalizeURL/relative_URL_with_leading_and_trailing_quotes-12      100000              2795 ns/op            1616 B/op         52 allocs/op
BenchmarkNormalizeURL/relative_URL_with_leading_and_trailing_quotes(OLD)-12 100000              1287 ns/op             368 B/op          7 allocs/op
BenchmarkNormalizeURL/relative_URL_without_parent-12                        100000              1351 ns/op             944 B/op         30 allocs/op
BenchmarkNormalizeURL/relative_URL_without_parent(OLD)-12                   100000              2263 ns/op             240 B/op          6 allocs/op
```

Unfortunately, in general, whatwg-url runs at 40-60% of the speed of goada on my i5-12450H laptap. (applied the code changes from my review.)

yzqzss:
I created a performance improvement PR for whatwg-url. https://github.com/nlnwa/whatwg-url/pull/44

After the optimization, the overall performance of whatwg-url should be 60%~75% of goada (the benchmark on https://github.com/ada-url/url-various-datasets/blob/main/top100/top100.txt is 67%), which is still not as good as goada, but not bad either.

anonrig:
Would you mind updating your benchmarks for Ada? I’m curious how is the new version…

otkd:
@yzqzss great points throughout, thank you!

yzqzss:
> Would you mind updating your benchmarks for Ada? I’m curious how is the new version…

Hmm, it appears to have no performance change.

internal/pkg/preprocessor/url.go
```diff
@@ -1,11 +1,18 @@
...
 
-	"github.com/ada-url/goada"
 	"github.com/internetarchive/Zeno/pkg/models"
+	wu "github.com/nlnwa/whatwg-url/url"
+)
+
+const (
+	httpPrefix  = "http://"
+	httpsPrefix = "https://"
+	ftpPrefix   = "ftp://"
```
yzqzss, line 15: Zeno don't support ftp yet, remove it?

yzqzss:
oh, whatwg-url allow us to use non-utf8 charsets to encode/decode URL path and query.
This is exactly what #370 needs!

yzqzss:
> oh, whatwg-url allow us to use non-utf8 charsets to encode/decode URL path and query. This is exactly what #370 needs!

It does support custom charsets, but only 8bit charsets...
and you can't use different custom encodings for path and query.

yzqzss:
Thanks for your efforts on this PR @otkd.

But I have to close this PR due to this � replacement issue, sorry :(

otkd:
> Thanks for your efforts on this PR @otkd.
> 
> But I have to close this PR due to this � replacement issue, sorry :(

@yzqzss all good, appreciate you trying to make it happen!

CorentinB:
This is interesting, @otkd would you mind reporting that whatwg-url bug to their repo?

willmhowes:
> I just made a CGO-free goada by using wazero. https://github.com/yzqzss/goada-wasm/
> 
> it works!
> image
> 
> Got 1x slow (compare to goada) on single-core, but almost no performence diffrence on multi-core.

@yzqzss Wow, incredibly interesting! As the resident goada defender, this is amazing!

However, I ask what problem we are trying to solve by removing cgo bindings in the first place? Originally it was to close https://github.com/internetarchive/Zeno/issues/245, but now I'm not sure?

yzqzss:
> Originally it was to close https://github.com/internetarchive/Zeno/issues/245, but now I'm not sure?

I have no idea, I saw this PR being reopened, so I tried packaging ada as WASM lib.

anonrig:
As one of the co-authors of Ada, I had an emotional roller coaster with this pull request lol

WASM will always be slower because of the cost of serializing string. I wrote the first version in Rust and compiled to WASM to be used by Node.js and saw similar performance characteristics. Gave a talk about this in case, it helps someone: https://youtu.be/v1lQK1jHZNQ

CorentinB:
My 2 cts is that any requirement of CGO in this project is more pain than needed, see: https://github.com/internetarchive/Zeno/issues/462

@willmhowes quite convince me yesterday by telling me that WASM is better than using the pure Go whatwg-url lib (if it was ever fixed) because we could be more up to date with the WASM route. So I think it would be a fine choice, even if my heart would still prefer to have as many things are possible in pure Go

PR #376, opened 2025-07-07
Implement `CheckDiskUsage()` for Windows

yzqzss:
#245

resolve: _Zeno's diskwatcher only uses linux syscall to get the free space size of the PWD._


NGTmeaty (approved):
I'm unable to test this on Windows right now, but it does look good! I'm fine with pushing this as is while allowing the test to fail. 

Let me know how that sounds.

PR #377, opened 2025-07-09
Implement a shared interface for `source` so anyone can implement their own source of items

equals215:
Most is said in the title.
What has been done :
- got rid of `source/hq` and `source/lq` global pointers
- `source/hq` and `source/lq` functions now work as methods and references are held if needed
- `preprocessor` now has a function to set a seencheck function, it is set at the `controler` and retains the same properties as before


NGTmeaty (approved):
Changes look good! Thank you for the PR!

PR #378, opened 2025-07-10
Fix outlinks extraction and widen HTML assets extraction

equals215:
1. When no outlinks extractor is used the code should continue trying with general-purpose extractors rather than returning without trying
2. Although it might not be HTML standard, `div` can hold a `data-src` or `data-srcset` and it should be captured


NGTmeaty (approved):
Looks good. Thanks again! 😄

PR #379, opened 2025-07-11
chore(deps): update github.com/ada-url/goada to v1.0.0

otkd:
Makes sense to bump in the interim even if we eventually end up proceeding with #374


NGTmeaty (approved):
Thanks!

PR #382, opened 2025-07-13
Simplify WARC writer stats

vbanos:
`internal/pkg/archiver/warc` had 8 functions to get 8 different stats. Each function iterated `[]*warc.CustomHTTPClient{globalArchiver.Client, globalArchiver.ClientWithProxy}` to get a stat.

We use all these functions just in one place except from one.

We can have a single `GetStats` function that returns a `WARCStats` struct with 8 integers. This, way, we'll do just one iteration. Also, now it is easy to extend with more stats.


NGTmeaty (approved):
Thanks!

PR #383, opened 2025-07-13
Fix: discard hooks does not interrupt downloads

yzqzss:
Also added some notes about discard hooks.


NGTmeaty (approved):
oops! Thanks for catching this and for the detailed notes on discard hooks!

PR #385, opened 2025-07-14
Fix: `BadStatusCode` responses are not handled by `copyWithTimeout()`

yzqzss:
bug: `--max-content-length` didn't work for those bad status responses before.

similar to #383


NGTmeaty (approved):
Thanks again!

PR #389, opened 2025-07-16
Refactor request preprocessors

vbanos:
Use a common Go interface. All preprocessors must implement `Match` and `Apply` methods. Rename the older arbitrary method names and use the interface.

Simplify calling sitespecific preprocessors to just `sitespecific.RunPreprocessors`.

Now, it is much simpler to define a new site specific preprocessor.

Add unit tests.

### DIFF
--- a/internal/pkg/preprocessor/sitespecific/sitespecific.go
+++ b/internal/pkg/preprocessor/sitespecific/sitespecific.go
@@ -1,0 +1,34 @@
+package sitespecific
+
+import (
+	"net/http"
+
+	"github.com/internetarchive/Zeno/internal/pkg/preprocessor/sitespecific/npr"
+	"github.com/internetarchive/Zeno/internal/pkg/preprocessor/sitespecific/reddit"
+	"github.com/internetarchive/Zeno/internal/pkg/preprocessor/sitespecific/tiktok"
+	"github.com/internetarchive/Zeno/internal/pkg/preprocessor/sitespecific/truthsocial"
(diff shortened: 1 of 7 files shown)


internal/pkg/preprocessor/sitespecific/sitespecific.go
```diff
@@ -0,0 +1,34 @@
...
+type Preprocessor interface {
+	Match(*models.URL) bool
+	Apply(*http.Request)
+}
+
+var preprocessors = []Preprocessor{
+	npr.NPRPreprocessor{},
+	reddit.RedditPreprocessor{},
+	tiktok.TikTokPreprocessor{},
+	truthsocial.TruthsocialStatusPreprocessor{},
```
NGTmeaty, line 22: Is TruthSocial intentionally in here twice after the checks were combined into one?

NGTmeaty: Oh, I see, `TruthsocialAccountsPreprocessor` is supposed to be there?

vbanos: Yes, this was a typo, I'm going to fix it after the meeting.
truthsocial has 2 preprocessors, one for the status and another for the account api.

vbanos: Done

NGTmeaty (approved):
Thank you!

PR #390, opened 2025-07-17
Refactor HTMLOutlinks to use models.URL instead of models.Item

vbanos:
We need all outlink extractors to use the same inputs / outputs for better code organization and correctness.
Without this change, we can't use Go interfaces.

Modify `HTMLOutlinks` extractor to use `models.URL` instead of `models.Item`.

Move the `base` attribute and related `GetBase`, `SetBase` methods from `Item` to `URL`. This also makes sense as the base URL attribute is more related to the `URL` object.

Modify `resolveURL` extractor to use `models.URL` instead of `models.Item`.

Modify `extractBaseTag` extractor to use `models.URL` instead of `models.Item`.

### DIFF
--- a/pkg/models/url.go
+++ b/pkg/models/url.go
@@ -21,6 +21,7 @@ type URL struct {
 	parsed    *url.URL
 	request   *http.Request
 	response  *http.Response
+    base       *url.URL     // Base is the base URL of the HTML doc, extracted from a <base> tag
 	body      spooledtempfile.ReadSeekCloser
 	document  *goquery.Document
 	mimetype  *mimetype.MIME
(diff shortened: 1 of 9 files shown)


pkg/models/url.go
```diff
@@ -21,6 +21,7 @@ type URL struct {
 	parsed    *url.URL
 	request   *http.Request
 	response  *http.Response
+    base       *url.URL     // Base is the base URL of the HTML doc, extracted from a <base> tag
```
NGTmeaty: Formatting here looks off

NGTmeaty (approved):
LGTM other than the one comment

pkg/models/url.go
vbanos: Fixed thank you.!

NGTmeaty (approved):

PR #391, opened 2025-07-17
Update ExtractAPIPostPermalinks to use models.URL instead of models.Item

vbanos:
We need all outlink extractors to use the same inputs / outputs for better code organization and correctness. Without this change, we can't use Go interfaces.

Modify ExtractAPIPostPermalinks extractor to use models.URL instead of models.Item.

Relevant to https://github.com/internetarchive/Zeno/pull/390


NGTmeaty (approved):
thanks!

PR #393, opened 2025-07-17
Move SetDocument(nil) to Item.Close

vbanos:
Placing `SetDocument(nil)` in the end of the method doesn't guarantee that it will be always executed. If the method returns early, it is skipped and memory isn't freed.

`Item` has a `Close()` method. Conceptually its better to include memory cleanup in `Item.Close`.

Also, method `postprocessItem` includes `defer item.Close()`.

So, it makes sense to move `SetDocument(nil)` in there.


yzqzss (approved):
LGTM

PR #400, opened 2025-07-18
Outlink extractor refactoring using interfaces

vbanos:
The "main" outlink extraction function `extractOutlinks` is reorganized in a major way.

All extractors are converted to use the same interface.

This makes it easier to add more outlink extractors in the future.


NGTmeaty (approved):
Looks great! Thank you for the refactor!

Issue #392, opened 2025-07-17
Double Hops increment for reddit.ExtractAPIPostPermalinks


vbanos:
`reddit.ExtractAPIPostPermalinks` increases the `Hops` of returned outlinks by 1

https://github.com/internetarchive/Zeno/blob/main/internal/pkg/postprocessor/sitespecific/reddit/api.go#L218

But in the end of `extractOutlinks` we also increment `Hops` for all found outlinks including the ones returned by `reddit.ExtractAPIPostPermalinks`

https://github.com/internetarchive/Zeno/blob/main/internal/pkg/postprocessor/outlinks.go#L98

Also, no other outlink extraction module increments the `Hops` of returned outlinks.

So, I think that we must remove `Hops +1` from `reddit.ExtractAPIPostPermalinks`.

PR #409, opened 2025-07-30
fix: double hops

yzqzss:
close: #392

This bug doesn't do any harm. The `extractOutlinks()` overrides Hops, so nothing happens.


NGTmeaty (approved):
Looks good! Thanks!

Issue #413, opened 2025-08-04
bufio.Scanner does not have large enough buffer for outlinks parsing


NGTmeaty:
`error scanning body for links component=postprocessor err=bufio.Scanner: token too long url=https://www.wikidata.org/wiki/Special:EntityPage/Q27536592`

It appears that the buffer is too small on the bufio.Scanner. 

This appears to be inconsistent but was triggered from:
```
./Zeno get url --headless --headless-headful --headless-dev-tools --headless-chromium-revision -1 --max-hops 2 "https://en.wikipedia.org/wiki/Hackmud"
```

PR #414, opened 2025-08-05
Optimize `extractLinksFromPage` by splitting words instead of lines

yzqzss:
fix: #413

---

Also made it a little faster.

```
before: BenchmarkExtractLinksFromPageRelax-12  465  2701707 ns/op  58202 KiB/s  369388 B/op     1311 allocs/op
after:  BenchmarkExtractLinksFromPageRelax-12  632  1941306 ns/op  80999 KiB/s  390980 B/op    15188 allocs/op
```


NGTmeaty (dismissed):
Looks great! Thank you for the quick changes! Do you think it would be worthwhile to add a test here?

NGTmeaty (approved):
Thanks!!

PR #415, opened 2025-08-05
New CSS parser with `go-css-lexer`

yzqzss:
https://github.com/renbaoshuo/go-css-lexer

This is a brand new CSS lexer, which complies with the latest CSS syntax.

The `go-css-lexer` can handle nested CSS, so in this PR I removed the previous regex fallback parser as well.

cc: @renbaoshuo

---

ref: https://github.com/internetarchive/Zeno/pull/345#issuecomment-2981327797


NGTmeaty (approved):
Looks great! I'm mostly just curious - but have you ran benchmarks on the old library vs the new one?

NGTmeaty (approved):
Thanks for the improvements there! That's great!

PR #417, opened 2025-08-08
Simplify HTML extractor logging

vbanos:
Define `url` and `item` once on logger init instead of calling them on each `Warn` and `Debug` call.


NGTmeaty (approved):
Thanks!

PR #418, opened 2025-08-08
Simplify Archiver logging

vbanos:
Define `seed_id` once on logger init instead of calling it on each logger call.


NGTmeaty (approved):
Thanks!

PR #419, opened 2025-08-08
Simplify Preprocessor logging

vbanos:
Define seed_id once on logger init instead of calling it on each logger call.

Also, there was a redundant `"func": "preprocessor.preprocess` which wasn't necessary as we already define that in the `component` field of the logger.


NGTmeaty (approved):
Thanks!

PR #420, opened 2025-08-08
Add max outlinks option

vbanos:
It is useful to be able to set a limit to the number of outlinks you extract.

Add relevant unit test.


NGTmeaty (approved):
Looks good! Thank you for the test!

PR #423, opened 2025-08-11
dependency: upgrade go-css-lexer to v0.1.0

yzqzss:


yzqzss:
```
csslexer v0.0.2 BenchmarkCSS      931164 ns/op             76681 kB/s   820866 B/op      15560 allocs/op
csslexer v0.0.3 BenchmarkCSS      538515 ns/op            132580 kB/s   447160 B/op         49 allocs/op
csslexer v0.1.0 BenchmarkCSS     1337031 ns/op             53389 kB/s   910266 B/op      23421 allocs/op
```

go-css-lexer has changed its internal behavior to convert the raw token ([]rune) into a token value (string) for each token. This makes it slower than the previous, unoptimized v0.0.2 for our use case.

The benefit, however, is that we can eliminate a lot of manual raw token parsing code.

What do you think?

CorentinB:
@yzqzss can you rebase so that we can merge? thx

CorentinB (approved):

PR #424, opened 2025-08-11
Simplify postprocessor logging

vbanos:
Define `item_id` once on logger init instead of calling it on each logger call.


NGTmeaty (dismissed):
thanks!

NGTmeaty (approved):
thanks!

PR #427, opened 2025-08-12
Use sync.Map in RateBucket instead of manual locking

vbanos:
Locking is slowing us down, this seems like a good case for `sync.Map`.


NGTmeaty (approved):
Looks good! Thank you!

PR #435, opened 2025-08-14
Create rootContext and share with components

vbanos:
Its not a good practice to create a separate context for each component without any special reasons.

We refactor to create a `rootContext` in `pipeline.Start` and pass it to many components.

One initial benefit of this approach is that you don't need to call `cancel()` in the "Stop" function of all components.
E.g. `consul.Stop()` is dropped because it isn't needed at all.

Now, `rootContextCancel()` is called in `pipeline.Stop` and all components using the `rootContext` are notified.

### DIFF
--- a/internal/pkg/consul/consul.go
+++ b/internal/pkg/consul/consul.go
@@ -13,16 +13,9 @@ import (
 	"github.com/internetarchive/Zeno/internal/pkg/log"
 )
 
-var ctx, cancel = context.WithCancel(context.Background())
-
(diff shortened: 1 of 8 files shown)


yzqzss:
I've noticed that using the shared global context causes components to exit in a random order, breaking the guarantees of stopPipeline() .
For example, the postprocessor almost always exits before the archiver, preventing items from being cleanly closed.

Do you mind if I revert this PR @vbanos ? I'm sorry I didn't award this earlier.

PR #442, opened 2025-08-19
More versatile HQ client init

vbanos:
Pass `HQKey, HQSecret, HQProject, HQAddress` as params on `HQ.New` instead of loading them via `config.Get()` inside `HQ.Start`.

This way, we can have multiple HQ client instances with different parameters.

This is necessary to be able to create a 2nd HQ client instance which connects an "outlinks" HQ project to send outlinks there.

### DIFF
--- a/internal/pkg/controler/pipeline.go
+++ b/internal/pkg/controler/pipeline.go
@@ -123,7 +123,7 @@ func startPipeline() {
 	finisherProduceChan := makeStageChannel(config.Get().WorkersCount)
 
 	if config.Get().UseHQ {
-		hqSource := hq.New()
+		hqSource := hq.New(config.Get().HQKey, config.Get().HQSecret, config.Get().HQProject, config.Get().HQAddress)
 		preprocessor.SetSeenchecker(hqSource.SeencheckItem)
 		sourceInterface = hqSource
 	} else {
--- a/internal/pkg/source/hq/hq.go
+++ b/internal/pkg/source/hq/hq.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"sync"
 
-	"github.com/internetarchive/Zeno/internal/pkg/config"
 	"github.com/internetarchive/Zeno/internal/pkg/log"
 	"github.com/internetarchive/Zeno/internal/pkg/reactor"
 	"github.com/internetarchive/Zeno/pkg/models"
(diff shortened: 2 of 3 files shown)


CorentinB:
Hi! How is that necessary compared to modifying HQ client to accept a new (optional) parameter (like HQOutlinksProject) that would be used instead of HQProject if specified?

vbanos:
I can explain my plan:

In `startPipeline()` we will add the following:
```
    hqOutlinksFinishChan := makeStageChannel(config.Get().WorkersCount)
    hqOutlinksProduceChan := makeStageChannel(config.Get().WorkersCount)

    // Optional 2nd HQ instance just to gather outlinks to a different project
    if config.Get().UseHQ && config.Get().HQOutlinksProject != "" && config.Get().HQOutlinksHopLimit > 0 {
        hqOutlinks := hq.New(config.Get().HQKey, config.Get().HQSecret, config.Get().HQOutlinksProject, config.Get().HQAddress)
        hqOutlinks.Start(hqOutlinksFinishChan, hqOutlinksProduceChan)
    }
```

Also, the post processor will change from `postprocessor.Start(inputChan, outputChan)` to `postprocessor.Start(inputChan, outputChan, hqOutlinksProduceChan)` to include the new channel that will send outlinks to the different project.

Then, in `postprocessor.go`, we will add the following function
```
 // If options UseHQ, HQOutlinks & HQOutlinksHopLimit are selected, send outlinks
// to a different HQ project and don't return them for further processing.
func (p *postprocessor) sendToHQOutlinks(outlinks []*models.Item) []*models.Item {
    if config.Get().UseHQ && config.Get().HQOutlinksProject != "" && config.Get().HQOutlinksHopLimit > 0 {
        var filtered []*models.Item
        for i := range outlinks {
            if outlinks[i].GetURL().GetHops() >= config.Get().HQOutlinksHopLimit {
                p.hqOutlinksProduceCh <- outlinks[i]
            } else {
                filtered = append(filtered, outlinks[i])
            }
        }
        return filtered
    } else {
        return outlinks
    }
}
```

And finally, in postprocessor `worker()` we will use it after we extract all outlinks
```
outlinks := postprocess(workerID, seed)
outlinks = p.sendToHQOutlinks(outlinks)
```

And we are done. The required changes are minimal.

IMHO the current hq pkg is very nice and well tested, we shouldn't touch it. We may introduce bugs.

If we tried adding `HQOutlinksProject` to `hq`, we would have to over-complicate it by adding a 2nd `client`, `wg`, `producer()` function for outlinks and even more (I haven't considered everything yet).

willmhowes:
@vbanos For the time being, let's proceed with your proposal that keeps the logic in Zeno. Given that HQ isn't open source at this time, I'd rather the functionality be in Zeno where all users can take advantage of it

PR #443, opened 2025-08-19
Periodic sqlite vacuuming for lq queue

yzqzss:
Execute `VACUUM;` every ten minutes to free LQ sqlite reusable space.

#94.

---
the branch is based on #441, pls merge it before.


CorentinB:
As I expected and as Copilot confirmed, VACUUM does block the operations of the DB, I don't think it should be used like that. Although maybe a trigger of VACUUM when shutting down Zeno? It might take a lot of time though so.. not sure. Maybe Copilot's suggestion is enough.

PR #449, opened 2025-08-26
Add --warc-digest-algorithm

CorentinB:
This pull request introduces support for selecting the digest algorithm used for WARC record digests, allowing users to choose between `sha1`, `sha256`, and `blake3`. It updates command-line flags, configuration, and internal logic to handle this new option, and ensures the selected algorithm is validated and passed through to the WARC writer. Additionally, it updates dependencies to support the new functionality.

**Digest Algorithm Selection and Validation:**

* Added a new persistent flag `--warc-digest-algorithm` (default: `sha1`) to the `get` command, allowing users to specify the digest algorithm for WARC records. Supported values are `sha1`, `sha256`, and `blake3`.
* Extended the `Config` struct to include a `WARCDigestAlgorithm` field, and added validation in `GenerateCrawlConfig()` to ensure the specified algorithm is supported. [[1]](diffhunk://#diff-3e598aea4c917756a4f220af33afe8cfb4b39dcd87ff264ff5a00f17686bcfb6R57) [[2]](diffhunk://#diff-3e598aea4c917756a4f220af33afe8cfb4b39dcd87ff264ff5a00f17686bcfb6R288-R292)
* Updated the WARC writer initialization to pass the selected digest algorithm to the underlying WARC library by calling `warc.GetDigestFromPrefix`.

**Dependency Updates:**

* Upgraded `github.com/internetarchive/gowarc` to v0.8.87 to support additional digest algorithms, and added indirect dependencies for `github.com/zeebo/blake3` and `github.com/klauspost/cpuid/v2`. [[1]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L21-R21) [[2]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R70) [[3]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R101)

**Minor Fixes:**

* Fixed a typo in the WARC rotator settings struct field name from `WarcSize` to `WARCSize`.
* Added import of the `gowarc` package in the config file to access digest-related helpers.

_**PS**: copilot wrote that, not that bad?_

PR #450, opened 2025-08-26
Switch default digest algorithm from SHA1 to BLAKE3

CorentinB:
## Why this change?
Zeno claims to be a **"state-of-the-art web crawler"**, but its current use of **SHA1** for payload digest calculation is outdated and problematic. SHA1 is cryptographically broken, as demonstrated by [the SHAttered attack](https://shattered.io/), which showed practical collision vulnerabilities. While SHA1 may still be used in some legacy contexts, **its use for deduplication in a modern crawler is unacceptable**, collisions can lead to data integrity issues and false deduplication.

## Why BLAKE3?
BLAKE3 is a *fast, secure, and modern alternative, offering both high performance and strong collision resistance. It is widely recommended for new systems:
- [BLAKE3 vs SHA2/SHA3: Performance and Security](https://kerkour.com/fast-secure-hash-function-sha256-sha512-sha3-blake3)
- [BLAKE2/BLAKE3 as High-Performance Alternatives](https://guptadeepak.com/blake2-and-blake3-high-performance-hashing-alternatives/)
- The web archiving community has also discussed moving away from SHA1 in favor of more robust algorithms ([IIPC #80](https://github.com/iipc/warc-specifications/issues/80)), and BLAKE3 has been mentioned multiple times and stands out as the best choice right now.

## Implementation Note
This PR depends on [#449](https://github.com/internetarchive/Zeno/pull/449), which introduces the `--warc-digest-algorithm` flag, allowing users to override the default if needed.

## Impact
- **Default changes from SHA1 to BLAKE3** for all new digests.
- **Backward compatibility**: new crawls will use BLAKE3 by default, but crawlers can still use SHA1 with `--warc-digest-algorithm sha1`, though this is **not recommended**.
- **Performance**: BLAKE3 is significantly faster than SHA1 while being more secure. (same goes for BLAKE3 vs SHA256 btw)

This change aligns Zeno with modern best practices and reinforces its claim as a state-of-the-art tool.

### DIFF
--- a/cmd/get.go
+++ b/cmd/get.go
@@ -114,7 +114,7 @@ func addWARCFlags(getCmd *cobra.Command) {
 	getCmd.PersistentFlags().String("warc-operator", "", "Contact informations of the crawl operator to write in the Warc-Info record in each WARC file.")
 	getCmd.PersistentFlags().String("warc-cdx-dedupe-server", "", "Identify the server to use CDX deduplication. This activates CDX deduplication.")
 	getCmd.PersistentFlags().String("warc-doppelganger-dedupe-server", "", "Identify the server to use doppelganger deduplication. This activates doppelganger deduplication.")
-	getCmd.PersistentFlags().String("warc-digest-algorithm", "blake3", "Digest algorithm to use for WARC records (Block & Payload digests). Possible values are: sha1, sha256, blake3. (note: SHA-1 is cryptographically broken and should not be used due to collision vulnerabilities)")
+	getCmd.PersistentFlags().String("warc-digest-algorithm", "sha256", "Digest algorithm to use for WARC records (Block & Payload digests). Possible values are: sha1, sha256, blake3. (note: SHA-1 is cryptographically broken and should not be used due to collision vulnerabilities)")
 	getCmd.PersistentFlags().Bool("warc-on-disk", false, "Do not use RAM to store payloads when recording traffic to WARCs, everything will happen on disk (usually used to reduce memory usage).")
 	getCmd.PersistentFlags().Int("warc-pool-size", 1, "Number of concurrent WARC files to write.")
 	getCmd.PersistentFlags().Int("warc-queue-size", -1, "Number of WARC records to queue before blocking the workers. Default is the --warc-pool-size.")


internal/pkg/archiver/warc.go
```diff
@@ -14,7 +14,7 @@ func startWARCWriter() {
 	rotatorSettings := warc.NewRotatorSettings()
 	rotatorSettings.Prefix = config.Get().WARCPrefix
 	rotatorSettings.WARCWriterPoolSize = config.Get().WARCPoolSize
-	rotatorSettings.WarcSize = float64(config.Get().WARCSize)
+	rotatorSettings.WARCSize = float64(config.Get().WARCSize)
```
willmhowes: Nice catch!

willmhowes (dismissed):
~lgtm~

EDIT: Referencing Jake's comment below (https://github.com/internetarchive/Zeno/pull/450#issuecomment-3235004573), I fear that I look like a doofus without the context of some conversations the maintainers had outside of this git issue. That being said, I agree with Jake's position below.

NGTmeaty:
From a purely WBM perspective, we currently require SHA1 to validate revisit records. Using BLAKE3 in revisit records simply won't be resolvable at this time. Because of that I don't think I can immediately recommend setting the default to BLAKE3. 

From a standards perspective, WARC producers often prefer standards stability and longevity over performance. One of the [main concerns around ZStandard](https://github.com/iipc/warc-specifications/issues/53#issuecomment-504650607) was the IETF RFC standard being finalized and widespread adoption. After 6 years both of those ended up happening! 
BLAKE3 has been out for a similar amount of time, but doesn't seem to have similarly wide adoption (and [the IETF draft](https://datatracker.ietf.org/doc/draft-aumasson-blake3/) is now expired.) I don't think either of these are disqualifying on their own but build into a larger picture and makes me question if it is the best path forward.

NGTmeaty:
I was incorrect on the point above about revisit, but the point still stands in regards to BLAKE3. I am torn on what our default should be, between SHA-256 and BLAKE3. 

As an absolute value right now, hashing is a relatively nominal 1.5% of overall CPU time. Even a minor increase would likely be okay, so I'm not sure speed should be one of our main factors.

PR #456, opened 2025-08-29
Add support for Sentry

### LINKED ISSUE #453: Pyroscope can only be used with HQ
Pyroscope can be enabled using `--pyroscope-address` both when running get url and get hq, but it's only applied in get hq, and nothing is logged when it's being used with get url. It should simply work with both.

NGTmeaty:
Also fixes pyroscope for URL-scoped crawl. Closes #453


willmhowes (approved):
lgtm!

PR #480, opened 2025-09-18
fix(docker): upgrades go version to 1.25.1

wilsoniya:
Upgrades version of Go used in Dockerfile, fixing a failure caused while building container images: 

```
[1/2] STEP 6/8: RUN go mod download
go: go.mod requires go >= 1.25.1 (running go 1.24.3; GOTOOLCHAIN=local)
subprocess exited with status 1
```


NGTmeaty (approved):
Thank you!

Issue #473, opened 2025-09-10
Excessive use of panic()


vbanos:
We use `panic()` 63 times(!) From my experience its not good to terminate the whole program in so many places.
I would expect to use `panic()` on the key components to terminate the program if they receive some kind of errors from their sub-components.
Maybe this is an indication that we don't do error handling correctly.

Also, in unit tests I would expect to use `t.Fatal` instead of `panic()`.

I will investigate more and get back with concrete improvement suggestions.

```
pkg/models/item_test.go
27:             panic(err)

pkg/models/url.go
110:            panic(err)

pkg/models/item.go
383:            panic("parent or child is nil")
419:                    panic(fmt.Sprintf("unable to close body, err: %s, seed id: %s", err.Error(), i.GetShortID()))

cmd/utils.go
48:             panic(fmt.Errorf("error starting pyroscope: %w", err))

internal/pkg/source/lq/consumer.go
101:                    panic(err)
145:                            panic("same seed received twice by lq.consumerSender")
178:                            panic(err)

internal/pkg/source/hq/consumer.go
104:                    panic(err)
110:                    panic(err)
158:                            panic("same seed received twice by hq.consumerSender")
191:                            panic(err)

internal/pkg/source/hq/seencheck.go
17:             panic(err)
49:             panic("no URLs to seencheck (can be caused if no fresh children were found)")

internal/pkg/preprocessor/preprocessor.go
90:             panic("preprocessor is not initialized")
94:             panic("seenchecker is already set")
133:                                    panic(fmt.Sprintf("seed consistency check failed with err: %s, seed id %s, worker_id %s", err.Error(), seed.GetShortID(), workerID))
137:                                    panic(fmt.Sprintf("preprocessor received seed with status %d, seed id: %s, worker_id %s", seed.GetStatus(), seed.GetShortID(), workerID))
164:            panic(err)
243:            panic(err)
263:            panic(err)

internal/pkg/preprocessor/seencheck/seencheck.go
38:             panic(err)
80:             panic(err)

internal/pkg/finisher/finisher.go
98:                                     panic("received nil seed")
102:                                    panic("received non-seed item")
108:                                    panic(fmt.Sprintf("seed consistency check failed with err: %s, seed id %s, worker id %s", err.Error(), seed.GetShortID(), workerID))
124:                                            panic(err)
133:                                    panic(err)

internal/pkg/reactor/reactor.go
99:             panic("item is not a seed")
136:                    panic("item is not a seed")
[...]

PR #481, opened 2025-09-19
Improve startPipeline error handling

vbanos:
Return error and use `panic` only once in the controller.

This way its cleaner and it would be easier to use it in tests.


NGTmeaty (approved):
Looks great. Thank you!

PR #482, opened 2025-09-19
Improve Archiver error reporting

vbanos:
Instead of crashing `startWARCWriter` via `os.Exit(1)`, we return the error which is passed via `archiver.Start` in the pipeline. There, the program is terminated via `panic` as it happens with all component errors.

### DIFF
--- a/internal/pkg/archiver/worker.go
+++ b/internal/pkg/archiver/worker.go
@@ -51,6 +51,8 @@ func Start(inputChan, outputChan chan *models.Item) error {
 		"component": "archiver",
 	})
 
+	var onceErr error
+
 	once.Do(func() {
 		ctx, cancel := context.WithCancel(context.Background())
 		globalArchiver = &archiver{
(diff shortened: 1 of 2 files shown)


internal/pkg/archiver/worker.go
```diff
@@ -76,7 +78,10 @@ func Start(inputChan, outputChan chan *models.Item) error {
 		logger.Debug("initialized")
 
 		// Setup WARC writing HTTP clients
-		startWARCWriter()
+		if err := startWARCWriter(); err != nil {
```
NGTmeaty, line 81: Would it make more sense to define this as WARCWriterErr or something similar?

vbanos: Now, we return the `err` returned by `warc.NewWARCWritingHTTPClient`.

How can we change the type of that to `WARCWriterErr` ?

NGTmeaty: I was just talking about the variable name itself not the type, but on second glance I think it's fine as is.

PR #483, opened 2025-09-19
Return errors in config init instead of os.Exit

vbanos:
Config already returns errors and terminates when there is an issue. There is no need to use `os.Exit(1)` for these specific cases, return error and the program will terminate anyway where `config.InitConfig` is invoked if it gets an error.

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -221,8 +221,7 @@ func InitConfig() error {
 		} else {
 			home, err := os.UserHomeDir()
 			if err != nil {
-				fmt.Println(err)
-				os.Exit(1)
+				return
 			}
 
 			viper.AddConfigPath(home)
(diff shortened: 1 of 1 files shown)


internal/pkg/config/config.go
```diff
@@ -221,8 +221,7 @@ func InitConfig() error {
 		} else {
 			home, err := os.UserHomeDir()
 			if err != nil {
-				fmt.Println(err)
-				os.Exit(1)
+				return
```
NGTmeaty, line 224: We should return the error here?

NGTmeaty (commented):
Looks good besides the one comment!

internal/pkg/config/config.go
vbanos: No, I can explain.
This code runs inside `once.Do(func() {` which means that we can only run `return`. It also happens in other places in the same function.

We define `err` here and then after we close `once.Do`, we return `err` if its not `nil`.

NGTmeaty: Ahhhh, I see. Thank you!

NGTmeaty (approved):
Thank you!

PR #484, opened 2025-09-19
Refactor item errors

vbanos:
Previously, when we had an item error we returned `fmt.Errorf("item is fresh but has children")`. and then in the unit tests we had to use
`expected: errors.New("item is fresh but has children"`

There is no need to define an error multiple times. We define:
```
ErrItemFreshHasChildren = errors.New("item is fresh but has children")
```
and then use `ErrItemFreshHasChildren` everywhere.

This way, we can easily change the error text in one place and avoid breaking the unit tests.

Also, we could avoid doing error string comparison and use `errors.Is` but we can't do it right now because the tests have some special error cases and its not straightforward.


NGTmeaty (approved):
Thank you!

PR #487, opened 2025-09-20
Expose --max-outlinks

yzqzss:


NGTmeaty (approved):
ooooops! Great catch! Thank you!

PR #490, opened 2025-09-20
Order local queue URLs by number of hops and timestamp

NGTmeaty:
This should allow us to get more predictable behavior out of the local queue.


CorentinB (dismissed):

willmhowes (dismissed):
Really great idea @NGTmeaty !

PR #511, opened 2025-10-28
Using goada-wasm if `CGO_ENABLE=0`

yzqzss:


yzqzss:
https://github.com/golang/go/issues/44695#issuecomment-973685193

CI fails because CGO_ENABLE="1" but we no longer use cgo.

NGTmeaty (approved):
So sorry for the delay here, been super busy! This is great! Thank you!

PR #512, opened 2025-10-28
fix: "error while reseting"

yzqzss:
Issue: the global ctx has been cancelled during the shutdown process, `ResetURL()` has no chance to be issued.


NGTmeaty (approved):
Appreciate it! Thank you!

Issue #521, opened 2025-11-19
Validate ItemFailed status


NGTmeaty:
We should validate that ItemFailed gets reset in HQ (and probably LQ as well).

PR #522, opened 2025-11-19
Seencheck improvements

NGTmeaty:
This should stop the panics from Seencheck failures as well as add retries. We do have additional work to do in #521 .


internal/pkg/preprocessor/preprocessor.go
```diff
@@ -253,10 +254,19 @@ func preprocess(workerID string, seed *models.Item) error {
...
+			if err == nil {
+				break
+			}
+			time.Sleep(1 * time.Second)
+		}
+
 		if err != nil {
 			logger.Error("unable to seencheck seed", "err", err.Error())
-			return err
+			seed.SetStatus(models.ItemFailed)
```
NGTmeaty: This will need review. I don't believe it works correctly.

PR #545, opened 2026-01-18
fix: connection is nil

yzqzss:
This bug was introduced in #356.

I guess that when I refactored two archiver.go files to achieve headless functionality, I somehow copied and pasted the resp.Body code from headless/archiver directly into general/archiver.

I'm not sure about the actual impact of this issue; could it lead to leaked connections?


NGTmeaty (approved):
Looks good! I'm not certain either but definitely a good catch!

PR #551, opened 2026-01-30
Refactor asset extractors

AltayAkkus:
https://github.com/internetarchive/Zeno/blob/aa1fbdb71fd2d5f0ee30509c08e2ec4d6bfe20a2/internal/pkg/postprocessor/assets.go#L24-L27

I started Refactoring, this is less straightforward than I thought.

The massive switch case in `assets.go` calls these extractors:
| Name         | Match method               | Input         | Output | Extract method              | Input         | Output                     |
|--------------|----------------------------|---------------|--------|-----------------------------|---------------|----------------------------|
| INAExtractor | ina.IsAPIURL               | item.GetURL() | bool   | ina.ExtractMedias           | item.GetURL() | assets, err                |
| TruthSocial  | truthsocial.NeedExtraction | item.GetURL() | bool   | truthsocial.ExtractAssets   | item          | assets, outlinks, err      |
| m3u8         | extractor.IsM3U8           | item.GetURL() | bool   | extractor.M3U8              | item.GetURL() | assets, err                |
| JSON         | extractor.IsJson           | item.GetURL() | bool   | extractor.JSON              | item.GetURL() | assets, outlinks, err      |
| XML          | extractor.IsXML            | item.GetURL() | bool   | extractor.XML               | item.GetURL() | assets, outlinks, err      |
| HTML         | extractor.IsHTML           | item.GetURL() | bool   | extractor.HTMLAssets        | item          | assets, err                |
| EmbeddedCSS  | extractor.IsEmbeddedCSS    | item          | bool   | extractor.ExtractFromURLCSS | item.GetURL() | assets, atImportLinks, err |

### INA Extraction

https://github.com/internetarchive/Zeno/blob/aa1fbdb71fd2d5f0ee30509c08e2ec4d6bfe20a2/internal/pkg/postprocessor/assets.go#L35-L48

Is this supposed to be a fallback (if the matched INA is not a JSON anymore, but a HTML)? 

### Interface design
I tried to take inspiration from 
https://github.com/internetarchive/Zeno/blob/aa1fbdb71fd2d5f0ee30509c08e2ec4d6bfe20a2/internal/pkg/preprocessor/sitespecific/sitespecific.go#L13-L34
and decided to use 
```go
type AssetExtractor interface {
	Match(*models.URL) bool
	Extract(*models.Item) (assets, outlinks []*models.URL, err error)
}

var assetExtractors = []AssetExtractor{
	ina.INAExtractor{},
	truthsocial.TruthsocialExtractor{},
	extractor.M3U8Extractor{},
	extractor.JSONExtractor{},
	extractor.XMLExtractor{},
	extractor.HTMLAssetsExtractor{},
}
```

This breaks support for the `EmbeddedCSSExtractor` (which requires the full `Item` in it's Match method), but it has some special logic anyway
https://github.com/internetarchive/Zeno/blob/aa1fbdb71fd2d5f0ee30509c08e2ec4d6bfe20a2/internal/pkg/postprocessor/assets.go#L79-L90

### Trump social
The RegEx for the `statuses` API URL has changed, I added that. But the rest of the JSON has also heavily changed, I don't know if this is a prioritized feature, will do this if I find time inbetween.


AltayAkkus:
> Any question?

I finished refactoring all asset extractors to the defined interface, I have a question regarding the overall logic of asset and outlink extraction:
The switch case currently matches URL, if the given extractor thinks it can extract assets and/or outlinks, the case will write into `assets` and `outlinks`, WITHOUT accumulating. So if two cases are met, the first one returns early, the second one will not be reached.

I currently just loop through all extractors, and if they match I append their newly found assets or outlinks to an array which holds all extracted assets/outlinks.

The comment in `assets.go` mentions priority
https://github.com/internetarchive/Zeno/blob/aa1fbdb71fd2d5f0ee30509c08e2ec4d6bfe20a2/internal/pkg/postprocessor/assets.go#L25-L26

Should we find the most specific/highest priority Asset extractor and exclusively use it's extracted assets/outlinks? (Loop through the ordered extractors, break out when matched) I think @vbanos is the original author of that comment

internal/pkg/postprocessor/assets.go
```diff
@@ -37,64 +38,28 @@ var assetExtractors = []AssetExtractor{
...
-		}
-
-		HTMLAssets, err := extractor.HTMLAssets(item)
-		if err != nil {
-			logger.Error("unable to extract assets", "err", err.Error())
+	for _, ext := range assetExtractors {
+		// heavy debug log calls, can be ommited when merged
+		logger.Debug("AssetExtractor Match call", "url", item.GetURL())
+		if ext.Match(item.GetURL()) {
+			logger.Debug("matched extractor", "extractor", fmt.Sprintf("%T", ext))
```
AltayAkkus: When I designed the AssetExtractor interface I did not add a method/attribute which holds it's identity (e.g. Name() -> "JSONExtractor").
It is only needed for debugging, thats why we got the fmt import just for this 👀

AltayAkkus:
I don't have a solid harness to test the asset extraction logic against a corpus of items.

I think Zeno in it's entirety would profit from a serialized data format for `models.item` which can loaded with 
```
// go:embed ina.item
var inaItem []byte
item, ok := testutil.HydrateItem(inaItem).(*models.Item)
```

Writing tests is currently a little bit of a hassle, you have to do the same bag of tricks again and again and again.

AltayAkkus:
> Yes, use the same logic as before. Extractors that appear earlier in the list have higher priority; once a match is found, `break`.
nice, thx for the clarification, implemented that in cc557c

AltayAkkus:
Converted to draft to avoid wasting Github Actions Runners resources (if the CI is configured like that)

PR #553, opened 2026-02-02
Fix pause controller race condition

### LINKED ISSUE #552: Race-condition in pause controller
I was working on #74 and added a `--pause-on-startup` flag which immediately paused via the Pause Controller on startup.
This did not impair the execution of Zeno in any way shape or form, since a Pause is only send to all subscribers.
Even if the `pauseManager`s `isPaused == true`, a new subscriber will not be paused.

AltayAkkus:
Fixes #552 

Added a test which tests if a new subscriber to a already `pauseManager` is paused upon subscription.

### DIFF
--- a/internal/pkg/controler/pause/pause.go
+++ b/internal/pkg/controler/pause/pause.go
@@ -39,7 +39,7 @@ func Subscribe() *ControlChans {
 		select {
 		case chans.PauseCh <- msg:
 		default:
-			// Should never happen since we just created the channel
+			// Channel already has a pause signal from concurrent Pause() call.
 		}
 	}
 


internal/pkg/controler/pause/pause.go
```diff
@@ -48,17 +64,20 @@ func Pause(message ...string) {
...
+	manager.message = msg
+	manager.mu.Unlock()
 
 	manager.subscribers.Range(func(key, _ any) bool {
 		chans := key.(*ControlChans)
-		// Send pause signal (non-blocking since PauseCh is buffered).
+		// Send pause signal with message (non-blocking since PauseCh is buffered).
 		select {
-		case chans.PauseCh <- struct{}{}:
+		case chans.PauseCh <- msg:
```
AltayAkkus, line 80: This is the biggest risk of incompatibility/breaking changes.

In the previous versio nof pause.go we just send struct{}{}, now we send the msg, I thought this was ok. lmk

NGTmeaty (approved):
This PR looks great to me besides the minor nits from Copilot! And I just wanted to say thanks for the other PRs so far!

AltayAkkus:
> This PR looks great to me besides the minor nits from Copilot! And I just wanted to say thanks for the other PRs so far!

Any timeline when this will be merged? thx in advance :)

Issue #74, opened 2024-07-09
Add basic UI to manage Zeno


CorentinB:
So the idea is basically to "replicate" the excellent Heritrix3 web UI.

We want to give a way to start, stop, pause, unpause the crawl, but also inject seeds, search in crawl logs, maybe remove stuff matching a query from the frontier.. The possibilities are endless.

![image](https://github.com/internetarchive/Zeno/assets/5089772/c995bf4a-e943-4d35-9bf2-5116184cf4c0)

Omramanuj:
hi, @CorentinB  is this still open?

CorentinB:
> hi, @CorentinB  is this still open? 

Hi, yes, but please work on dev/v2 not the main branch :)

Omramanuj:
I'm facing some challenges with routing to serve an HTML page and also encountering disk space filling out after some run. Understanding the codebase from someone experienced would be really helpful. Is there a Slack or Discord channel where I can ask such queries?

Any guidance would be greatly appreciated. Thanks!

CorentinB:
@Omramanuj sadly we don't have such thing, but I invite you to use the Discussions feature & issues/PRs :)

Omramanuj:
> [@Omramanuj](https://github.com/Omramanuj) sadly we don't have such thing, but I invite you to use the Discussions feature & issues/PRs :)

Thanks, @CorentinB! I found a way to handle it and will finish with the code soon. For future queries, I'll definitely use Discussions & issues/PRs. Appreciate the guidance!

PR #556, opened 2026-02-03
Feature: Web UI

AltayAkkus:
#74 proposed adding a UI to manage Zeno, I came up with the following:
# General architecture
My proposal is to **heavily extend the already existing API** (currently only used by Prometheus), breaking out as many functions as possible.

1. The API shall act as a **general control interface**, not only for our Web UI but also for use-cases like **automations**

(Example given: A seed was unsuccessful, we did not find any other pages. A script automatically performs a Google dork search `site:example.com`, and adds the newly found URLs back to the reactor. Or you bruteforce known paths, or use the robots.txt + bruteforce).

2. The **Web UI shall be interchangeable**, because every org has different requirements and likes other flavors.

The API can now be configured via `api-static-dir` to serve a local directory, you can input the `dist` from Vite, or write barebones `.html` yourself.

3. The API and WebUI shall allow **easy, authenticated access**, without turning Zeno into half-webserver half-webcrawler

All API endpoints (and the served files) are unauthenticated, and you can install a authenticator-proxy in front of it 
(I used Cloudflare Tunnels, the daemon can be installed on Linux, and you can configure access via SSO, via password or via whitelisted e-mails. You can also use AWS Cognito with CloudFront, or whatever floats your boat)

I propose avoiding RBAC, SSO et cetera to keep this thing as simple as possible, with an API + directory file server everybody can get this up and running fast.

## API feature list (to be extended)

- [x] Pause/Unpause
Implemented and working, I had to fix a bug in a55719 (the pause would not stop the crawling, only if a new seed begun)

- [x] Live-tailed Frontier
You can stream the Reactor StateTable via WebSocket now 🎉
There is a poller which polls the Reactor, computes delta, and fan-out's the StateTable in **realtime**. More details below.
- [ ] Add seeds to Reactor
- [ ] Live-tail logs
Similiar to the Frontier (websocket), conserving the full schema of logs (level and fields) so we can filter through the logs.

- [ ] WebHook
If you could register a WebHook on e.g. ERROR level logs, you could hook that up to Slack and get notifications whenever there is something wrong with your Zeno instance.

### DIFF
--- a/cmd/get.go
+++ b/cmd/get.go
@@ -53,6 +53,7 @@ func addBasicFlags(getCmd *cobra.Command) {
 	getCmd.PersistentFlags().Bool("api", false, "Enable API")
 	getCmd.PersistentFlags().Int("api-port", 9090, "Port to listen on for the API.")
 	getCmd.PersistentFlags().String("api-static-dir", "static", "Directory which shall be served as static files at /")
+	getCmd.PersistentFlags().Duration("api-frontier-poll-interval", time.Second, "Interval at which the frontier WebSocket polls the reactor for item changes. (Default: 1s)")
 	getCmd.PersistentFlags().Int("max-redirect", 20, "Specifies the maximum number of redirections to follow for a resource.")
 	getCmd.PersistentFlags().Int("max-css-jump", 10, "Specifies the maximum number of CSS @import jumps to follow for a resource.")
 	getCmd.PersistentFlags().Int("max-retry", 5, "Number of retry if error happen when executing HTTP request.")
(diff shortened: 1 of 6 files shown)


AltayAkkus:
# Frontier errata
I thought that just returning the StateTable via HTTP will lead to performance issues, especially when you want high-fidelity data
image

## Caveats
The streamed items can be inconsistent depending on when you opened that WebSocket channel (referencing parent IDs which were fan-out before you started your WS). For now this is an necessary evil, if anyone has a nice idea on how we could solve this I would be really glad :) 

The delta is calculated on item-level only, normally triggering a retransmit whenever the status of the Item changes. You could do this finer, to reduce the bandwidth of each WS.

We should also use something more efficient than JSON over-the-wire, [CBOR](https://cbor.io/) seems like a good choice.

# Demo

https://github.com/user-attachments/assets/b4ba22e0-7155-42b1-82ef-f6f413f3425a

PR #557, opened 2026-02-06
Add ItemFailed status when seencheck fails rather than panic

NGTmeaty:
We currently don't believe seencheck failing should result in a panic. This causes an incredible amount of pain when we can fail that specific item after 5 attempts.


internal/pkg/source/hq/seencheck.go
```diff
@@ -20,34 +21,54 @@ func (s *HQ) SeencheckItem(item *models.Item) error {
...
 
+	// If all URLs were resolved from cache, nothing to send to HQ
 	if len(URLsToSeencheck) == 0 {
-		panic("no URLs to seencheck (can be caused if no fresh children were found)")
+		logger.Debug("all URLs resolved from seencheck cache (or otherwise), no HQ request needed")
+		return nil
 	}
 
-	// Debug print the seencheck response
+	// Debug print the seencheck request
```
willmhowes, line 71: Nice catch!

willmhowes (approved):
LGTM!

PR #558, opened 2026-02-08
Detect and ignore path loops

NGTmeaty:
Detect and stop loops in path and query.

### DIFF
--- a/internal/pkg/preprocessor/pathloop.go
+++ b/internal/pkg/preprocessor/pathloop.go
@@ -27,6 +27,9 @@ import (
 func hasPathLoop(path, search string) bool {
 	maxRepetitions := config.Get().MaxSegmentRepetition
 	if maxRepetitions <= 0 {
+		if logger != nil {
+			logger.Info("maxRepetition unset - being reset to 3")
+		}
 		maxRepetitions = 3
 	}
 	// Check path segments
@@ -50,6 +53,9 @@ func hasPathLoop(path, search string) bool {
 	// even if no single segment exceeds maxRepetitions.
 	threshold := config.Get().MaxSegmentRepetitionThreshold
 	if threshold <= 0 {
+		if logger != nil {
+			logger.Info("maxSegmentRepetitionThreshold unset - being reset to 3")
+		}
 		threshold = 2
 	}
 	if nonEmptySegments >= 10 {


internal/pkg/preprocessor/url_cgofree.go
willmhowes, line 1: TIL we maintain a second url preprocessor that doesn't use goada

internal/pkg/preprocessor/pathloop_test.go
willmhowes, line 1: Tests should be modified to take an argument specifying `maxRepetitions`
EDIT: Only if we go forward with parameterizing `maxRepetitions`, and same would true for the tests in `url_test.go`

e2e/test/nxdomain/nxdomain_test.go
```diff
@@ -17,7 +17,7 @@ type recordMatcher struct {
 
 func (rm *recordMatcher) Match(record map[string]string) {
 	if record["level"] == "ERROR" {
-		if strings.Contains(record["err"], "no suitable IP address found") {
+		if strings.Contains(record["err"], "failed to resolve DNS") {
```
willmhowes, line 20: Bonus bug fix? 😄

internal/pkg/preprocessor/url_cgofree.go
willmhowes: I suggest we find a way to dynamically set the `goada` package imported by a single `url.go` file rather than maintaining two separate files with one difference

internal/pkg/preprocessor/pathloop.go
```diff
@@ -0,0 +1,73 @@
...
+
+import "strings"
+
+// maxRepetitions is the repetition threshold used to detect crawler traps.
+// A URL is considered a trap when:
+//   - any single path segment or query parameter key=value pair appears
+//     more than maxRepetitions times, OR
+//   - the path has 10+ segments and at least 2 distinct segments each
+//     appear maxRepetitions or more times (the deep-path heuristic).
+const maxRepetitions = 3
```
willmhowes: I'd like this to be configurable at runtime, thoughts?

internal/pkg/preprocessor/pathloop.go
```diff
@@ -0,0 +1,73 @@
+package preprocessor
+
+import "strings"
+
+// maxRepetitions is the repetition threshold used to detect crawler traps.
```
willmhowes: I would add "non-consecutive" to this line explaining `maxRepetitions`

internal/pkg/preprocessor/pathloop.go
```diff
@@ -0,0 +1,73 @@
...
+	// Deep-path heuristic: in paths with 10+ segments, flag when 2 or more
+	// distinct segments each appear at least maxRepetitions times (>=, not >).
+	// This catches complex traps where several segments repeat together
+	// even if no single segment exceeds maxRepetitions.
+	if nonEmptySegments >= 10 {
+		segmentsAtThreshold := 0
+		for _, count := range counts {
+			if count >= maxRepetitions {
+				segmentsAtThreshold++
+				if segmentsAtThreshold >= 2 {
```
willmhowes: Can we move `2` to its own var at top of file so it's not a magic number?

willmhowes: And if we're parameterizing `maxRepetitions`, then _maybe_ this as well?

e2e/test/nxdomain/nxdomain_test.go
NGTmeaty: yes :)

internal/pkg/preprocessor/url_cgofree.go
NGTmeaty: Well, so both are goada, one is `goada-wasm` (so no CGO needed). My recollection is that it was rather annoying to deal with so this was the easiest solution (and not a huge deal to maintain). If you disagree, I'm happy to look into it in another issue.

willmhowes (approved):

PR #563, opened 2026-02-22
Feat: Item test fixture

AltayAkkus:
When you want to create a `models.Item` for tests
previously
https://github.com/internetarchive/Zeno/blob/d475a600b41a5de342e00ea0aeaa14cee47d2c93/internal/pkg/postprocessor/extractor/html_test.go#L15-L56
now
```golang
//go:embed testdata/ina_api_response.json
var inaFixture []byte

func TestExtractAssets_HydrateItemFixture(t *testing.T) {
	item := testutil.HydrateItem(t, inaFixture)
	assets, _, err := ExtractAssetsOutlinks(item)
         ...
}
```

You can create the JSON files by hand (look at the `ina_api_response.json` for ereference), define everything you need.
Or you can create them using this [curl drop-in alias](https://gist.github.com/AltayAkkus/3f77606dcf784c43b894b4fbb1cd66f5) which just spits the JSON out.


AltayAkkus:
Merged into #551

Issue #569, opened 2026-03-02
Strict regex performance issue


yzqzss:
pre: #290 

https://github.com/mvdan/xurls/blob/ee9499d3fc7b6557f435e241350b262c5d1ded82/xurls.go#L145

I found that `xurl.Strict()` trying to match various schemes, the schemes list is so long that we saw very huge performance downgreade with --strict-regex enabled.

PR #570, opened 2026-03-02
fix: strict regex performance issue

yzqzss:
close: #569

fixed by limiting the schemes to http and https. (btw, xurls is case-insensitive by default)

---

image


yzqzss:
@copilot 

1dc39ece648ced0a34ec3c341f7794e963773b4a

Please find out what URL is missing from RSS test due to this PR change, and whether it is a valid URL. Thx.

PR #574, opened 2026-03-03
Archiver: fix race condition, dead code path, and remove spin-wait

shramanb113:
## Summary

This PR improves the reliability and scalability of the archiver by removing
fragile timing assumptions, eliminating dead code execution, and replacing
a spin-wait in the rate limiter.

## Changes

1. Replaces a `time.Sleep` used to avoid a known race condition between
`router.Run()` and `page.Navigate()` with a deterministic synchronization
mechanism.

2. When asset capture and crawling are disabled, the response body was fully
discarded but processing continued, leading to MIME detection and buffering
on an already-consumed stream. An early return is added to avoid unnecessary
work and ensure correct behavior.

 3. Replaces a busy-wait loop that periodically re-acquired a mutex with a
time-based wait calculated  from the token refill rate.


shramanb113:
Hi @yzqzss ,
I saw that this PR was marked as AI-generated. I want to make it clear that I actually wrote these changes myself. I spent some time debugging. Made these changes by hand. I'm not super familiar with the codebase. I was able to find the issues with the ratelimiter because I've implemented similar things in my own projects.

The other changes I made were just adding comments here and there. I did this because using time.sleep to run the server isn't always the approach.

I promise you I put in effort to make these changes. I think they will make the codebase more robust. If you need information, about how I did it I'd be happy to explain.

pkg/models/item_dedupe.go
```diff
@@ -38,10 +38,10 @@ func flattenTree(root *Item) []*Item {
 	var nodes []*Item
 	var traverse func(node *Item)
 	traverse = func(node *Item) {
-		nodes = append(nodes, node)
 		if node == nil {
 			return
 		}
+		nodes = append(nodes, node)
```
NGTmeaty, line 44: This appears to be from your other closed PR. Could we get this removed?

internal/pkg/archiver/ratelimiter/ratelimiter.go
yzqzss, line 1: re-enter dead lock:

lock() -> unlock() -> lock() -> (continue) -> lock(!)

internal/pkg/archiver/headless/archiver.go
yzqzss, line 1: I think this is not the correct fix?

PR #596, opened 2026-05-17
feat: ensure URL length does not exceed reasonable limits

NGTmeaty:


CorentinB:
See, I leave the Internet Archive and a year later I see PRs with "reasonable" in the title.. 😛

internal/pkg/preprocessor/error.go
```diff
@@ -11,4 +11,6 @@ var (
 	ErrUnsupportedHost = errors.New("unsupported host")
 	// ErrPathLoopDetected is the error returned when a URL path contains repeating segments indicating a crawler trap
 	ErrPathLoopDetected = errors.New("URL path or query loop detected")
+	// ErrURLTooLong is the error returned when a URL exceeds the maximum allowed length
+	ErrURLTooLong = errors.New("URL exceeds maximum length of 4000 characters")
```
willmhowes: `max-url-length` is variable, so shouldn't reference the default value of 4000

internal/pkg/preprocessor/url_cgo.go
```diff
@@ -72,6 +73,10 @@ func NormalizeURL(URL *models.URL, parentURL *models.URL) (err error) {
 	pathname := adaParse.Pathname()
 	search := adaParse.Search()
 
+	if maxLen := config.Get().MaxURLLength; maxLen > 0 && len(href) > maxLen {
```
willmhowes: I'd personally remove the maxLen > 0 conditional

willmhowes (changes requested):

willmhowes (approved):
LGTM!

